PR #4, opened 2024-06-03
Improve CI

steve-downey:
Improve the CI workflow, building with clang 17, 18, and 19, against libc++, and  with GCC  13 and 14 using Ubuntu 24.04 as the base OS.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -85,11 +85,6 @@ jobs:
           chmod +x llvm.sh
           sudo ./llvm.sh ${{matrix.config.clang_version}} all
 
-      - name: Install GCC 13
-        if: matrix.config.name == 'Ubuntu GCC 13'
-        run: |
-          sudo apt update
-          sudo apt-get install g++-13
 
       - name: Install GCC 14
         if: matrix.config.name == 'Ubuntu GCC 14'
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,6 +17,15 @@ jobs:
       matrix:
         config:
         - {
+          name: "Ubuntu Clang 17",
+          os: ubuntu-24.04,
+          toolchain: "clang-17-toolchain.cmake",
+          clang_version: 17,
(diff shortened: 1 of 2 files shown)


neatudarius (approved):

etc/clang-flags.cmake
```diff
@@ -3,7 +3,7 @@ include_guard(GLOBAL)
 set(CMAKE_CXX_STANDARD 23)
```
JeffGarland, line 3: remind me -- is there something here that requires 23, or could it be 20?

.github/workflows/ci.yml
```diff
@@ -1,76 +1,121 @@
-name: ci
+name: Test
```
JeffGarland, line 1: I'm no expert on these github workflows now - @bretbrownjr any comments?  Specifically I'm wondering about non-Linux platforms.  Also, @steve-downey did this come from our template or do we need to update?

JeffGarland (approved):

.github/workflows/ci.yml
neatudarius: @steve-downey , is there a way to run these flows on local environment? e.g. git API to span a Docker for every job, but on my local setup. I am just curious if I am doing a change, if I can run all jobs on my MAC OS / Ubuntu.

steve-downey: I have not found a way to simulate any of this locally. I've ended up enabling actions on PRs and repeated force-pushes to a branch to trigger a run.

steve-downey: I took a bunch of it from https://github.com/beman-project/example/pull/6. I also have some possible updates to deal with Node version warnings over here: https://github.com/steve-downey/lambda-fun/pull/2/files

etc/clang-flags.cmake
steve-downey: I'll check. I believe at one point I was looking at `deducing this` but then didn't.

steve-downey:
Merging so as to base further work on working CI

PR #5, opened 2024-06-11
Configure Beman.Optional26

neatudarius:
* Update root README.md
* Update CODEOWNERS / .gitignore
* Remove unused build scripts & Docker infrastructure ("bld" is unused after #4). ~~Note: This PR will be merged after #4 - required new CI rules to pass if `bld` is removed.~~
* Move some left over paper build related files from root to paper subdirectory.

This partially solves #12 .

PR #6, opened 2024-06-12
Apply the Beman Standard; more tree refactoring

neatudarius:
Apply the [Beman Standard](https://github.com/beman-project/beman/blob/main/docs/beman-standard.md) to C++ code & more tree refactoring:

* Move `optional.hpp` to `include/Beman/Optional26/optional.hpp`; rename `src/beman` to `src/Beman` to match the `include/` tree.
* Rename static library from `optional.a` to `beman_optional26`.
* Move/add examples to `/examples`,

This should solve remaining parts from #12, after #5.


src/beman/optional/optional_monadic.t.cpp
```diff
@@ -1,6 +1,4 @@
-#include <beman/optional/optional.hpp>
-
-#include <beman/optional/optional.hpp>
+#include <Beman/Optional26/optional.hpp>
```
neatudarius, line 1: No need for second include here, we already test this in `src/beman/optional/optional.t.cpp`.

src/beman/optional/optional_monadic.t.cpp
```diff
@@ -85,27 +83,26 @@ TEST(OptionalMonadicTest, Transform) {
 
   }
 
-TEST(OptionalMonadicTest, TransformConstexpr) {
-
```
neatudarius, line 89: Some linting issues were also fixed.

steve-downey (approved):
LGTM 👍

PR #7, opened 2024-06-12
[WIP] CI checks improvement: add clang-format check

neatudarius:


.github/workflows/ci.yml
```diff
@@ -17,30 +37,30 @@ jobs:
       matrix:
         config:
         - {
-          name: "Ubuntu Clang 17",
```
neatudarius, line 20: For consistency with other blocks in the file.

.github/workflows/ci.yml
```diff
@@ -7,6 +7,26 @@ on:
     branches: [ main ]
 
 jobs:
+  lint:
```
steve-downey, line 10: This https://cpp-linter.github.io/cpp-linter-action/ might be a friendlier linter interface?

.github/workflows/ci.yml
```diff
@@ -7,6 +7,26 @@ on:
...
+    name: Lint C++
+    runs-on: ubuntu-24.04
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
```
steve-downey, line 15: v3 is deprecated, switch to v4 to fix the warnings about node deprecation.

.github/workflows/ci.yml
```diff
@@ -7,6 +7,26 @@ on:
...
+
+      - name: Install Clang Format
+        run: |
+          sudo apt-get update
+          sudo apt-get install clang-format
```
steve-downey, line 20: clang-format gets stale fast for contemporary C++, we should make sure we're getting a current one.

steve-downey (commented):
Notes
Generally looks good. 
Do you have a yaml formatter that is good? The indent I missed in the config blocks irks me.

.github/workflows/ci.yml
neatudarius: I tried your suggestion. I couldn't make it work and also it uses clang-12 https://github.com/beman-project/Optional26/actions/runs/9494415629/job/26164716972 :( .


I was also thinking about an market Git action, but then I realized this doesn't help me to apply changes locally. Do you think it's better to have a script (e.g., `lint-all.sh` - cpp, sh, md, yaml etc) and be able to run locally and from github workflow?

steve-downey: I'm talking with some other developers locally about how they handle this for public github projects. We have a much nicer internal workflow tool, of course. 
The weak consensus seems to be that for clang-format it's hard to justify a separate action that needs to build a docker image, and it's better to get the clang-format that you want, and then apply something like https://github.com/bloomberg/blazingmq/blob/d3bf9637db34a91455c9d83354c4844c178c3f1e/.github/workflows/formatting-check.yaml#L22
```
        run: |
          git clang-format-14 --diff -q origin/main | tee format_diff.txt
          if [ -s format_diff.txt ]; then exit 1; fi
```          

Which is hard coded to clang-format-14 which is so ancient it may not understand C++ anymore. But we've already got some infrastructure for getting LLVM current added. 

clang-tidy may be more complicated, since it often needs review to confirm applicability? One suggestion has been
https://github.com/marketplace/actions/clang-tidy-review

Which does allow current clang-tidy to be used. 

Making anything we do in automated checks something that can be straightforwardly done locally is important to me, as well. There's little as frustrating as finding out 30 minutes later that you've done something that would have been trivial to fix as you were coding it. Shifting left in the business process vernacular.

neatudarius:
>  Do you have a yaml formatter that is good? The indent I missed in the config blocks irks me.

Check experimental commit 21a53d4347a6e112015b3fcb30c400bacc847c7f

neatudarius:
work moved in #26

PR #9, opened 2024-06-13
Implement P3168R1: Give std::optional Range Support

neatudarius:
Implement [P3168R1: Give std::optional Range Support](https://wg21.link/P3168R1) - issue #14 .

* Apply P3168R1 wording
* Apply P3168R1 wording also for optional<T&>
* Move tests into a subdirectory
* Add tests for features related to P3168R1.
* Temporary disable clang-19 cI flow until we can locally replicate it on Ubuntu x86_64 and arm64 (check: https://github.com/llvm/llvm-project/issues/95380).

Note: `__cpp_lib_format_ranges`  available only on `clang` (check comments from `optional.hpp`.

PR #15, opened 2024-06-18
Debug clang-19 failure

steve-downey:


JeffGarland:
Seems like clang 17 isn't quite up to the task, do we really care? Seems likely that the ongoing 'what compilers' to support discussion will continue to heat up as we do real work.

steve-downey:
It is bothering me because it appears to work on my machine. But I think
that is because I happen to have the libc++ from 19 installed, and that
avoids the particular error.
Libc++ from 17 did not have join, so sone tests fail.
It may not be worth going back to clang-17, though, I agree.

neatudarius:
@steve-downey , I merged #9 . You can update your branch to see why the clang-19 job failed on that particular job/code.

steve-downey:
I'll put up a cleaner PR now that this is working.

Issue #17, opened 2024-06-19
Use std::iterator_interface from P2727 as an iterator type for Beman.Optional26


neatudarius:
As discussed in a previous Beman sync meeting (06-17-2024), as a long-term approach, we want to use std::iterator_interface from https://wg21.link/P2727R4 into the Beman project, and thus, in Beman.Optional26.

Note: Check[ Boost implementation](https://github.com/boostorg/stl_interfaces).

neatudarius:
Link against https://github.com/beman-project/beman/issues/36

neatudarius:
@bretbrownjr , what's the current status of the packaging initiative?

I would like to known when it's stable and we can try to use beman.iterator_interface as a dependency for beman.optional. Thanks!

Issue #14, opened 2024-06-13
Implement proposal P3168: Give std::optional Range Support


neatudarius:
Implement proposal [P3168: Give std::optional Range Support](https://wg21.link/P3168R1), currently forwarded from LEWG to LWG according to [P3214R0 2024-04 Library Evolution Poll Outcomes](https://wg21.link/P3214R0). 

* Current revision:  [https://wg21.link/P3168R1](https://wg21.link/P3168R1)
* [LEWG issue](https://github.com/cplusplus/papers/issues/1831)

- [x] Implement P3168R1
- [x] Demo on Compiler Explorer
- [x] Implement P3168R2
- [x] Add usage examples for P3168

PR #18, opened 2024-06-19
P3168R1: use a non-pointer iterator type

neatudarius:
P3168R1: use a non-pointer iterator type . Follow-up PR after #9 for suggestion made by @brevzin (issue #14) .

Updates:
* `include/Beman/Optional26/detail/iterator.hpp`: Adopt [normal_iterator](https://github.com/gcc-mirror/gcc/blob/master/libstdc%2B%2B-v3/include/bits/stl_iterator.h#L1030-L1338) implementation from GCC and create a simplified version into `beman::optional::detail:normal_iterator`.
* `include/Beman/Optional26/optional.hpp`: Replace implementation for `optional<T/T&>::iterator/const_iterator`: `T*/const*` -> `detail::normal_iterator<T/const T*, optional>` .
* Add more tests related to iterators.

~~Notes:~~
* ~~This should be a very small PR (interesting diff in `optional.hpp`, also possible in `iterator.hpp`).~~
* ~~A long-term implementation for the iterator type will be done in #17. Now we just want something good enough to be able to make experiments for LWG@Saint Louis.~~
L.E.
As discussed in [this thread](https://github.com/beman-project/Optional26/pull/18#discussion_r1647675872), I copied the stl_interfaces Boost implementation into our tree (#17).

### DIFF
--- a/include/Beman/Optional26/detail/__iterator.hpp
+++ b/include/Beman/Optional26/detail/__iterator.hpp
@@ -1,0 +1,147 @@
+// include/Beman/Optional26/detail/__iterator.hpp                                            -*-C++-*-
+#ifndef BEMAN_OPTIONAL26_OPTIONAL_DETAIL___ITERATOR_HPP
+#define BEMAN_OPTIONAL26_OPTIONAL_DETAIL___ITERATOR_HPP
+
+#include <iterator>
+#include <type_traits>
+
+namespace beman::optional {
+// Implementation inspired by libstdc++'s normal_iterator
+// (https://github.com/gcc-mirror/gcc/blob/master/libstdc%2B%2B-v3/include/bits/stl_iterator.h)
+//
+//
+// This iterator adapter is a normal iterator in the sense that it does not
+// change the semantics of any of the operators of its iterator
+// parameter.  Its primary purpose is to convert an iterator that is
+// not a class, e.g. a pointer, into an iterator that is a class.
+//
+// The ContainerType parameter exists solely so that different containers
+// using this template can instantiate different types, even if the
+// IteratorType parameter is the same.
+//
+// FUTURE Use std::iterator_interface from https://wg21.link/P2727R4.
+// Adopt Boost's iterator_interface (https://github.com/boostorg/stl_interfaces) into a Beman library.
+//
+template <typename IteratorType, typename ContainerType>
+class normal_iterator {
+  public:
+    using iterator_type     = IteratorType;
+    using iterator_category = std::iterator_traits<iterator_type>::iterator_category;
+    using iterator_concept  = std::iterator_traits<iterator_type>::iterator_concept;
+    using value_type        = std::iterator_traits<iterator_type>::value_type;
+    using difference_type   = std::iterator_traits<iterator_type>::difference_type;
+    using reference         = std::iterator_traits<iterator_type>::reference;
+    using pointer           = std::iterator_traits<iterator_type>::pointer;
+
+    // Default constructor.
+    constexpr normal_iterator() noexcept : m_current(IteratorType()) {}
+
+    // Constructor from an iterator.
+    explicit constexpr normal_iterator(const IteratorType& iter) noexcept : m_current(iter) {}
+
+    // Allow iterator to const_iterator conversion.
+    template <typename OtherIteratorType,
+              typename = std::enable_if_t<std::is_convertible_v<OtherIteratorType, IteratorType>>>
+    constexpr normal_iterator(const normal_iterator<OtherIteratorType, ContainerType>& iter) noexcept
+        : m_current(iter.base()) {}
+
+    constexpr const auto& base() const noexcept { return m_current; }
+
+    // Forward iterator requirements
+    constexpr reference operator*() const noexcept { return *m_current; }
+
+    constexpr pointer operator->() const noexcept { return m_current; }
+
+    constexpr normal_iterator& operator++() noexcept {
+        ++m_current;
+        return *this;
+    }
+
+    constexpr normal_iterator operator++(int) noexcept { return normal_iterator(m_current++); }
+
+    // Bidirectional iterator requirements
+    constexpr normal_iterator& operator--() noexcept {
+        --m_current;
+        return *this;
+    }
+
+    constexpr normal_iterator operator--(int) noexcept { return normal_iterator(m_current--); }
+
+    // Random access iterator requirements
+    constexpr reference operator[](difference_type n) const noexcept { return m_current[n]; }
+
+    constexpr normal_iterator& operator+=(difference_type n) noexcept {
+        m_current += n;
+        return *this;
+    }
+
+    constexpr normal_iterator operator+(difference_type n) const noexcept { return normal_iterator(m_current + n); }
+
+    constexpr normal_iterator& operator-=(difference_type n) noexcept {
+        m_current -= n;
+        return *this;
+    }
+
+    constexpr normal_iterator operator-(difference_type n) const noexcept { return normal_iterator(m_current - n); }
+
+  protected:
+    // The underlying pointer iterator.
+    IteratorType m_current;
+};
+
+// Forward iterator requirements
+template <typename IteratorTypeL, typename IteratorTypeR, typename ContainerType>
+[[nodiscard]] constexpr inline bool operator==(const normal_iterator<IteratorTypeL, ContainerType>& lhs,
+                                               const normal_iterator<IteratorTypeR, ContainerType>& rhs) noexcept {
+    return lhs.base() == rhs.base();
+}
+
+template <typename IteratorTypeL, typename IteratorTypeR, typename ContainerType>
+[[nodiscard]] constexpr inline bool operator!=(const normal_iterator<IteratorTypeL, ContainerType>& lhs,
+                                               const normal_iterator<IteratorTypeR, ContainerType>& rhs) noexcept {
+    return lhs.base() != rhs.base();
+}
+
+// Random access iterator requirements
+template <typename IteratorTypeL, typename IteratorTypeR, typename ContainerType>
+[[nodiscard]] inline bool operator<(const normal_iterator<IteratorTypeL, ContainerType>& lhs,
+                                    const normal_iterator<IteratorTypeR, ContainerType>& rhs) noexcept {
+    return lhs.base() < rhs.base();
+}
+
+template <typename IteratorTypeL, typename IteratorTypeR, typename ContainerType>
+[[nodiscard]] inline bool operator>(const normal_iterator<IteratorTypeL, ContainerType>& lhs,
+                                    const normal_iterator<IteratorTypeR, ContainerType>& rhs) noexcept {
+    return lhs.base() > rhs.base();
+}
+
+template <typename IteratorTypeL, typename IteratorTypeR, typename ContainerType>
+[[nodiscard]] inline bool operator<=(const normal_iterator<IteratorTypeL, ContainerType>& lhs,
+                                     const normal_iterator<IteratorTypeR, ContainerType>& rhs) noexcept {
+    return lhs.base() <= rhs.base();
+}
+
+template <typename IteratorTypeL, typename IteratorTypeR, typename ContainerType>
+[[nodiscard]] inline bool operator>=(const normal_iterator<IteratorTypeL, ContainerType>& lhs,
+                                     const normal_iterator<IteratorTypeR, ContainerType>& rhs) noexcept {
+    return lhs.base() >= rhs.base();
+}
+
+// According to the resolution of DR179 not only the various comparison
+// operators but also operator- must accept mixed iterator/const_iterator
+template <typename IteratorTypeL, typename IteratorTypeR, typename ContainerType>
+[[nodiscard]] constexpr inline auto operator-(const normal_iterator<IteratorTypeL, ContainerType>& lhs,
+                                              const normal_iterator<IteratorTypeR, ContainerType>& rhs) noexcept {
+    return lhs.base() - rhs.base();
+}
+
+template <typename IteratorType, typename ContainerType>
+[[nodiscard]] constexpr inline normal_iterator<IteratorType, ContainerType>
+operator+(typename normal_iterator<IteratorType, ContainerType>::difference_type n,
+          const normal_iterator<IteratorType, ContainerType>&                    iter) noexcept {
+    return normal_iterator(iter.base() + n);
+}
+
+} // namespace beman::optional
+
+#endif // BEMAN_OPTIONAL26_OPTIONAL_DETAIL___ITERATOR_HPP
--- a/include/Beman/Optional26/detail/iterator.hpp
+++ b/include/Beman/Optional26/detail/iterator.hpp
@@ -2,30 +2,28 @@
 #ifndef BEMAN_OPTIONAL26_OPTIONAL_DETAIL_ITERATOR_HPP
 #define BEMAN_OPTIONAL26_OPTIONAL_DETAIL_ITERATOR_HPP
 
+#include <Beman/Optional26/detail/stl_interfaces/iterator_interface.hpp>
+
 #include <iterator>
-#include <type_traits>
 
 namespace beman::optional::detail {
-// Implementation inspired by libstdc++'s normal_iterator
-// (https://github.com/gcc-mirror/gcc/blob/master/libstdc%2B%2B-v3/include/bits/stl_iterator.h)
+
+// This is a minimal contiguous iterator.  It uses stl_interfaces library from Boost
+// (current implementation based on https://wg21.link/P2727R4).
 //
+// TODO: Change this to use the stl_interfaces library from Beman when available.
 //
-// This iterator adapter is a normal iterator in the sense that it does not
-// change the semantics of any of the operators of its iterator
-// parameter.  Its primary purpose is to convert an iterator that is
-// not a class, e.g. a pointer, into an iterator that is a class.
-//
-// The ContainerType parameter exists solely so that different containers
-// using this template can instantiate different types, even if the
-// IteratorType parameter is the same.
-//
-// FUTURE Use std::iterator_interface from https://wg21.link/P2727R4.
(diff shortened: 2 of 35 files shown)


src/Beman/Optional26/tests/__utils.h
```diff
@@ -0,0 +1,36 @@
+#ifndef BEMAN_OPTIONAL26_OPTIONAL_TESTS__UTILS_HPP
```
steve-downey, line 1: utils is a sure to collide name for things, and the names of types in here are, too. This should be beman/optional/test/utils.hpp or something similar? 
Namespace shouldn't be just ::test, or when this is composed into a larger group it's going to collide when linked, even if no TU ends up with two "::test::no_default_ctor" types.

src/Beman/Optional26/tests/optional_range_support.t.cpp
```diff
@@ -213,7 +180,7 @@ TEST(RangeSupportTest, LoopOverEmptyRange) {
     beman::optional::optional<int> empty;
     ASSERT_FALSE(empty.has_value());
 
-    for (auto _ : empty) {
+    for ([[maybe_unused]] auto _ : empty) {
```
steve-downey, line 183: Does this make the compiler less unhappy about _ not being used?

include/Beman/Optional26/detail/__iterator.hpp
```diff
@@ -0,0 +1,147 @@
+// include/Beman/Optional26/detail/__iterator.hpp                                            -*-C++-*-
+#ifndef BEMAN_OPTIONAL26_OPTIONAL_DETAIL___ITERATOR_HPP
+#define BEMAN_OPTIONAL26_OPTIONAL_DETAIL___ITERATOR_HPP
+
+#include <iterator>
+#include <type_traits>
+
+namespace beman::optional {
```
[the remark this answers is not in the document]
neatudarius: Will change to `detail/iterator.cpp`.

src/Beman/Optional26/tests/detail/__iterator.t.cpp
```diff
@@ -0,0 +1,176 @@
+#include <Beman/Optional26/detail/__iterator.hpp>
+
+/**
+ * This file contains tests for normal_iterator.
+ */
+#include <gtest/gtest.h>
+#include "../__utils.h"
```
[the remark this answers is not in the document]
neatudarius: done

src/Beman/Optional26/tests/optional_range_support.t.cpp
neatudarius: At least for some compilers :D

include/Beman/Optional26/detail/iterator.hpp
```diff
@@ -0,0 +1,147 @@
+// include/Beman/Optional26/detail/iterator.hpp                                            -*-C++-*-
+#ifndef BEMAN_OPTIONAL26_OPTIONAL_DETAIL_ITERATOR_HPP
+#define BEMAN_OPTIONAL26_OPTIONAL_DETAIL_ITERATOR_HPP
+
+#include <iterator>
+#include <type_traits>
+
+namespace beman::optional::detail {
+// Implementation inspired by libstdc++'s normal_iterator
+// (https://github.com/gcc-mirror/gcc/blob/master/libstdc%2B%2B-v3/include/bits/stl_iterator.h)
```
camio, line 10: Some of this code looks exactly like libstdc++'s class. libstdc++'s copyright and license should be reflected in this file.

src/Beman/Optional26/tests/detail/iterator.t.cpp
```diff
@@ -0,0 +1,178 @@
+/**
```
camio, line 1: This file needs license information per [FILE.LICENSE_ID].

include/Beman/Optional26/detail/iterator.hpp
camio: Hrm, and libstdc++'s license is not a Beman approved license.

JeffGarland: I would suggest using boost.iterator or stl_interfaces (this needs c++20 I believe. https://github.com/boostorg/stl_interfaces

Note that stl_interfaces is proposed for c++26 as well so this would be a good use case for it.

src/Beman/Optional26/tests/detail/iterator.t.cpp
```diff
@@ -0,0 +1,178 @@
...
+ * This file contains tests for beman::optional::detail::normal_iterator.
+ */
```
JeffGarland, line 3: Why do we have a tests/detail here -- shouldn't it just be tests?  I don't think we need to 'hide' some sort of implementation details for test code.

steve-downey: If we're not putting the tests in the directory with the code, putting them in a mechanically derivable location makes it easier / possible to teach an IDE or editor to open related files. 
This is a lot of ceremony that seems unneeded on a project with one real file, but with larger projects knowing where things are and go makes it much easier to navigate.

steve-downey: The library is currently building with std 23. I think it would probably work with 20. Deducing this is the "killer feature" for  libraries in 23, but it's not used here. 

Writing in 20 style makes translation to and from standard wording *very* straightforward.

JeffGarland: I'm saying this test file (and all) should be in Optional26/tests -- NOT Optional26/tests/detail. I'm not following why we need that extra directory level and then basically leaving tests empty?

neatudarius: I prefer to have a tree identical for production code and tests code.

include/Beman/Optional26/detail/iterator.hpp
neatudarius: @JeffGarland , pleaze check issue linked in the PR description

neatudarius: Ok. I offered to help @tzlaine to important his iterator to Beman. I understood the wording may change.

JeffGarland: Well I would hope the wording would be the same, but don't stop!

src/Beman/Optional26/tests/detail/iterator.t.cpp
JeffGarland: Ok well I consider it strange. What I was missing is that already on main is the public interface tests.

neatudarius: yes, I merged it in a previous PR with initial implementation using T*

neatudarius: I propose to move this discussion into a Beman Standard thread. To be honest, I'm happy to change it to any convention if we have a guideline.

include/Beman/Optional26/detail/iterator.hpp
neatudarius: @camio , @JeffGarland , I made an experiment directly using `boost/stl_interfaces` as an external dep to `Beman.Optional26`. I was able to build and test with clang-18 (not sure yet why gcc13/14 won't work). 

Diff can be seen in https://github.com/neatudarius/Beman.Optional26/pull/2/.

Is this direction OK for everyody?

camio: Hrm, it'd be great to not pull in a huge Boost dependency. Is it feasible to put a copy of that code in the detail namespace instead?

neatudarius: I copied 3 files from that Boost library in a6c6fa3bae9367733ebf354f4b7ae24c7da6965b

```shell
tree include/Beman/Optional26/detail 
include/Beman/Optional26/detail
├── iterator.hpp
└── stl_interfaces
    ├── config.hpp
    ├── fwd.hpp
    └── iterator_interface.hpp
 ```
 
 Next commit (f4eaa880ea6397907dc3636f215c69dc3b2c7189) renames the namespace / macros to be specific to `beman::optional::detail`.
 
 @camio , can you take a look?

File: src/Beman/Optional26/tests/detail/iterator.t.cpp (hunk 6 above)
Comment (neatudarius) [reply src/Beman/Optional26/tests/detail/iterator.t.cpp:4]: Please check - updated entire tree in 416ec70bd1e9da43702b29f4c01a231d22c2069d

File: include/Beman/Optional26/detail/iterator.hpp (hunk 5 above)
Comment (JeffGarland) [reply include/Beman/Optional26/detail/iterator.hpp]: The better convention here is probably to put the stl_interfaces under external in the repository, but this is probably fine as well.  Ideally any macros / namespaces needed in stl interfaces would go back into stl_interfaces primary repo so that in the end a fetch content cmake would be able to grab the dependency instead of copying here.  I took a brief look at the iterator and it looked fine, but will try to dig deeper later.

So you're spearheading a number of technical things for Beman here -- I expect to spend some time on Thursday night discussing this project at the Beman night session.  Good stuff :)

Comment (neatudarius) [reply include/Beman/Optional26/detail/iterator.hpp]: We discussed 2 approaches in this thread:

1. I've initially attached a commit from other branch were I had `external/stl_interfaces` - check 7457c76f9254b8ea494c3d6e790865c15f54b2d2. In this approach, I tried not change at all the Boost code (used `extern/stl_interfaces/include` as include directory in `src/Beman/Optional26/CMakeLists.txt`, and was able to directly include headers from `boost/stl_interfaces` in `iterator.hpp` in this repo).  

2. @camio  suggested that we better copy 3 headers from `boost/stl_interfaces` in `include/Beman/Optional26/detail` and then we rename whatever means boost names (namespaces, macros etc) to be sure that we don't have any interaction with other boost library versions installed on target system - basically, make `Beman.Optionak26` a standalone library, no production deps. When we'll have a `Beman.STLInterfaces` library available we can remove copied files.

> The better convention here is probably to put the stl_interfaces under external in the repository, but this is probably fine as well. Ideally any macros / namespaces needed in stl interfaces would go back into stl_interfaces primary repo so that in the end a fetch content cmake would be able to grab the dependency instead of copying here. 

@JeffGarland , I have a clarifying question here: are you aiming for option1 or option 2 or a mixt of them/new option 3? The `fetch content` suggestion seems to be `option 2` with no extra changes :D

Comment (JeffGarland) [reply include/Beman/Optional26/detail/iterator.hpp]: For now, I'm basically fine with either -- but I think our long term target should be fetch content with Beman stuff incorporated into the Boost source repo. Unless it is @tzlaine vision to port stl_interfaces to Beman specific repo -- but I don't see that as being needed. I very much prefer the external location (copied or fetched) since that makes it clear that this is work from elsewhere that is technically incidental to the implementation here. I certainly expect this won't be our last Boost dependency -- still many useful tools over there for library building.

Comment (neatudarius) [reply include/Beman/Optional26/detail/iterator.hpp]: > So you're spearheading a number of technical things for Beman here -- I expect to spend some time on Thursday night discussing this project at the Beman night session. Good stuff :)

Do we have something on agenda? :D please share a link.

Comment (JeffGarland) [reply include/Beman/Optional26/detail/iterator.hpp]: https://wiki.edg.com/bin/view/Wg21stlouis2024/EveningSessions

Are you going to be there?

Comment (neatudarius) [reply include/Beman/Optional26/detail/iterator.hpp]: ok. that's 3:30 AM Bucharest time :( I'll try...

File: include/Beman/Optional26/detail/stl_interfaces/iterator_interface.hpp [hunk 8]
```diff
@@ -0,0 +1,986 @@
+// include/Beman/Optional26/detail/stl_interfaces/iterator_interface.hpp -*-C++-*-
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
```
camio, line 2: This should be the Boost license identifier. Also, the top-level README should mention that some code is licensed under Boost and will be marked as such.

camio: Well, unless you could get @tzlaine to give you an "Apache-2.0 WITH LLVM-exception" license to his code. That would make it simpler.

steve-downey:
I had a little time this weekend:
https://github.com/steve-downey/iterator-interface
It's an implementation from the wording and exposition from P2727R4 - Only
"creative" bit was writing the iterator tag calculation.

Did turn up some wording nits, so not totally useless.

include/Beman/Optional26/detail/stl_interfaces/iterator_interface.hpp
tzlaine: This is my official permission to use the Boost.STLInterfaces code under the Apache-2.0 WITH LLVM-exception license.

neatudarius: That's great, @tzlaine! One more step closer to have a demo for Beman sync in Saint Louis.

Also, please check #21 (local demo with Godbolt and my fork of Beman.Optional26).

neatudarius:
> I had a little time this weekend: https://github.com/steve-downey/iterator-interface It's an implementation from the wording and exposition from P2727R4 - Only "creative" bit was writing the iterator tag calculation. Did turn up some wording nits, so not totally useless.

Nice! Thanks Steve! I think we can start a chat with @tzlaine and see that's the plan for stl_interfaces inside Beman.

include/Beman/Optional26/detail/iterator.hpp
neatudarius: Closed this thread as now we have a copy of boost/stl_interfaces sources + permission to use them under "Apache-2.0 WITH LLVM-exception"  licence (https://github.com/beman-project/Optional26/pull/18#discussion_r1651196265).

include/Beman/Optional26/detail/iterator.hpp
```diff
@@ -0,0 +1,73 @@
...
+//
+// @tparam T - The type of the elements the iterator points to.
+// @tparam Container - The type of the container the iterator points to. This parameter exists solely so that different
+// containers using this template can instantiate different types, even if the T parameter is the same.
+template <class T, class Container>
+struct contiguous_iterator : public base_contiguous_iterator<T, Container> {
+    // Alias for the base class.
+    using base_type = base_contiguous_iterator<T, Container>;
+    // Alias for types from the base class.
+    using typename base_type::difference_type;
```
tzlaine, line 41: It looks like you need difference_type and pointer, but I don't think you need the rest.

tzlaine (approved):
I added a small note, but it's entirely style.  The rest looks good!

include/Beman/Optional26/detail/iterator.hpp
neatudarius: right, also need reference. removed unused aliases

include/Beman/Optional26/detail/iterator.hpp
```diff
@@ -59,6 +63,11 @@ struct contiguous_iterator : public base_contiguous_iterator<T, Container> {
     T* m_current;
 };
 
+// Check concepts for the contiguous_iterator class.
```
steve-downey, line 63: Move to a test. It's small in the scheme of things, but we don't need to check this every time this header is used.

steve-downey (changes requested):

include/Beman/Optional26/detail/iterator.hpp
neatudarius: We already have extensive tests for these in `src/Beman/Optional26/tests/detail/iterator.t.cpp`:

```shell
    // non-const iterators
    test(beman::optional::detail::contiguous_iterator<int, container>{});
    test(beman::optional::detail::contiguous_iterator<empty, container>{});
    test(beman::optional::detail::contiguous_iterator<no_default_ctor, container>{});
    test(beman::optional::detail::contiguous_iterator<base, container>{});
    test(beman::optional::detail::contiguous_iterator<derived, container>{});

    // const iterators
    test(beman::optional::detail::contiguous_iterator<const int, container>{});
    test(beman::optional::detail::contiguous_iterator<const empty, container>{});
    test(beman::optional::detail::contiguous_iterator<const no_default_ctor, container>{});
    test(beman::optional::detail::contiguous_iterator<const base, container>{});
    test(beman::optional::detail::contiguous_iterator<const derived, container>{});
 ```

I removed this check from the header.

File: include/Beman/Optional26/optional.hpp [hunk 11]
```diff
@@ -321,6 +322,9 @@ class optional {
     using iterator       = detail::contiguous_iterator<T, optional>;       // see [optional.iterators]
     using const_iterator = detail::contiguous_iterator<const T, optional>; // see [optional.iterators]
 
+    static_assert(std::contiguous_iterator<iterator>);
```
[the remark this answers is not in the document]
neatudarius: also here: we already have tests in `Beman/Beman.Optional26/src/Beman/Optional26/tests/optional_range_support.t.cpp`:

```shell
TEST(RangeSupportTest, IteratorConcepts) {
    const auto test = [](auto&& opt) {
        // The iterator type is the same as the iterator of the optional.
        // e.g. iterator = optional<T>::iterator if opt is optional<T>
        using iterator = typename std::remove_reference_t<decltype(opt)>::iterator;
        // e.g. const_iterator = optional<T>::const_iterator if opt is optional<T>
        using const_iterator = typename std::remove_reference_t<decltype(opt)>::const_iterator;

        ...
        static_assert(std::contiguous_iterator<iterator>);

        ...
        static_assert(std::contiguous_iterator<const_iterator>);
    };

    test(beman::optional::optional<int>{});
    test(beman::optional::optional<int*>{});
    test(beman::optional::optional<empty>{});
    test(beman::optional::optional<no_default_ctor>{});
    test(beman::optional::optional<base>{});
    test(beman::optional::optional<derived>{});
}
```

steve-downey (approved):

JeffGarland (approved):

camio (approved):

neatudarius:
Merged the PR to be able to deploy into Compiler Explorer (#21 ).

Follow-up tweaks will be done in other PRs.

PR #20, opened 2024-06-20
Update CODEOWNERS

neatudarius:
Noticed that only Steve was required as  codeowner due to wrong syntax in the file.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,5 +1,3 @@
 # Codeowners for reviews on PRs
 
-*       @camio
-*       @neatudarius
-*       @steve-downey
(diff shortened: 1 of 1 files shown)


.github/CODEOWNERS
```diff
@@ -1,5 +1,3 @@
 # Codeowners for reviews on PRs
 
-*       @camio
-*       @neatudarius
-*       @steve-downey
+*       @camio, @neatudarius, @steve-downey
```
neatudarius, line 3:
```suggestion
*       @camio @neatudarius @steve-downey
```

steve-downey (approved):
LGTM

Issue #16, opened 2024-06-19
Apply and document the minimum C++ Standard needed to compile Beman.Optional26


neatudarius:
Apply and document the minimum C++ Standard needed to compile Beman.Optional26

- [x] Experimental set of compilers and C++ standards 
- [x] Update docs

neatudarius:
Implemented in #23 and #26 .

Issue #11, opened 2024-06-13
Improve CI and local flows


neatudarius:
Linters should be checked and run on CI.

I propose to add linters for C++, shell and Markdown.

- [x] Linters should be checked and run on CI
- [x] Linters can be checked and run on local system, using the same scripts from CI.
- [x] Build and runt tests should be pulled into a generic script, which can be used in CI and local flows.

neatudarius:
WIP PR https://github.com/beman-project/Optional26/pull/26

neatudarius:
All subtasks very deployed. Closing this issue.

Issue #2, opened 2024-06-02
Optional.h missing 


steve-downey:
Lost in the combination of directory and file extension rename. Need to do some git surgery to make sure history is connected. Go back, create a branch, do the move and rename, and merge that so the object graph is preserved.

steve-downey:
#3 Fixes the  problem.

steve-downey:
PR merged and issue fixed. At least for me.

Issue #1, opened 2024-06-02
CI needs to be engaged. 


steve-downey:
Looks like the main branch isn't what is configured. Also need to set up multiple compilers and OS.

steve-downey:
fixed

PR #26, opened 2024-06-28
Improve local and CI flows

neatudarius:
TLDR: scripts for local and CI flows. #1 #2 #11 #16
The codebase (e.g. C++ files) are not updated in this PR. I will make a new PR an update `main` and rebase this branch until this PR becomes green!

===========================================================================

Updates already applied:

* Update toolchain files (`etc/` tree)
    * Remove unused toolchain files from `etc/` . Add license. Run linters.
    * Expose `CMAKE_CXX_STANDARD` (tested with C++20, C++23 and C++26)
   
 * Configure linters:
    * C++: `clang-format`
        * pulled version: `19`, tested on Ubuntu 24.04
        * config: `.clang-format`
    * Shell: `shell-check`
       * pulled version: default apt-get version on Ubuntu 24.04
       * config: `.shellcheckrc`   
    * Markdown: `markdownlint`
       * pulled version: default version available with `npm` on Ubuntu 24.04
       * config: `.markdownlint.json`
    *  YAML: `yamllint`
       *  pulled version: default apt-get version, on Ubuntu 24.04
       * config: `yamllint` 
    *  CMake: `cmake-format`
        * pulled version: default apt-get version, on Ubuntu 24.04
        * config: `.cmake-format`

* Scripts: Uniform testing and linting local and CI environments! Full usage in [scripts/README.md](https://github.com/neatudarius/Beman.Optional26/blob/improve-flows/scripts/README.md).
   *  `scripts/lint.sh`: Script used to lint some or all files in the project.
      * Can detect errors. Can automatically fix detected errors (`--fix`).
      
   *  `scripts/run-tests.sh`: Script used to run tests on local and CI environments!
      * Can use a single or multiple toolchains (`--preset`). Can set a C++ standard (`--std`).
      
      
* New CI flows:
   * `github/workflows/ci_tests.yml`: use `scripts/run-test.sh` and a matrix config to run all tests.
      * toolchans:  `{gcc-13, gcc-14, clang-17, clang-18, clang-19}`. Note: Expected gcc-13 to fail with C++26.
      * stds: `{20, 23, 26}`
  

   * `github/workflows/ci_lint.yml`: use `scripts/lint.sh` to make a PR have failed jobs if linting fails.
      * types of checks:  `{--cpp, --shell, --markdown,  --yaml, --cmake}`
      * examples of failed jobs: 
         * [clang-format failed job](https://github.com/beman-project/Optional26/actions/runs/9714431177/job/26813460163?pr=26)
         * [shellcheck failed job](https://github.com/beman-project/Optional26/actions/runs/9714431177/job/26813460951?pr=26)
         * [markdownlint failed job](https://github.com/beman-project/Optional26/actions/runs/9714431177/job/26813461332?pr=26)
         * [yamllint failed jobs](https://github.com/beman-project/Optional26/actions/runs/9714431177/job/26813461952?pr=26)
         * [cmake-format failed job](https://github.com/beman-project/Optional26/actions/runs/9714431177/job/26813462296?pr=26)
  
image image
          

===========================================================================

Next steps:
* All jobs from `ci_tests.yml` are green and can be deployed in `main`.
* Intentionally, all jobs from `ci_lint.yml` are red.  I will push start making commits to fix linter errors, after we agree a direciton.


README.md
```diff
@@ -4,7 +4,8 @@
 SPDX-License-Identifier: 2.0 license with LLVM exceptions
 -->
 
-![CI Tests](https://github.com/beman-project/Optional26/actions/workflows/ci.yml/badge.svg)
+![CI Tests](https://github.com/beman-project/Optional26/actions/workflows/ci_tests.yml/badge.svg)
```
neatudarius, line 7: New badges

.clang-format
```diff
@@ -1,5 +1,11 @@
----
+# /.clang-format -*-config-*-
```
steve-downey, line 1: This will make it slightly harder to keep the .clang-format up to date because clang-format will want to rewrite it when you dump the config. 
I believe it's right now at clang-format-17.
```sh
clang-format-19 --style=file:.clang-format    --dump-config
```
will produce the file that clang-format believes it read, adding the options that were defaulted.

.markdownlint.json
```diff
@@ -1,4 +1,12 @@
+// /.markdownlint.json -*-config-*-
```
steve-downey, line 1: Looks like github thinks that json comments are an error. 
That's annoying but getting all this red is also?

.clang-format
```diff
@@ -164,7 +170,7 @@ PenaltyBreakTemplateDeclaration: 10
 PenaltyExcessCharacter: 1000000
 PenaltyIndentedWhitespace: 0
 PenaltyReturnTypeOnItsOwnLine: 60
-PointerAlignment: Left
+# PointerAlignment: Left
```
steve-downey, line 173: This causes all the files to be marked as misformatted right now. If we change it, we should also reformat. But probably as a separate change that we can tell github to ignore in blame.

etc/clang-17-toolchain.cmake
```diff
@@ -1,3 +1,8 @@
+# cmake-format: off
```
steve-downey, line 1: Does cmake-format not like the file?

etc/clang-flags.cmake
```diff
@@ -1,13 +1,30 @@
+# cmake-format: off
+# etc/clang-flags.cmake -*-makefile-*-
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+# cmake-format: on
+
 include_guard(GLOBAL)
 
-set(CMAKE_CXX_STANDARD 23)
+if(NOT "${CMAKE_CXX_STANDARD}")
```
steve-downey, line 8: I'll work on fixing this up, but Standard level is one of the things that needs to be a fixed part of a toolchain so that ABI is consistent across components built with the same toolchain, it's not just about which compiler. 
We're building everything in tree and not depending on anything that has polyfills so it's not going to cause a problem for optional26, but it does in general. In particular once modules are in play.

etc/gcc-13-toolchain.cmake
```diff
@@ -1,3 +1,8 @@
+# cmake-format: off
+# etc/gcc-13-toolchain.cmake -*-makefile-*-
```
steve-downey, line 2: `-*- cmake -*- ` unless you want makefile-mode for editing cmake files.

scripts/README.md
```diff
@@ -0,0 +1,324 @@
+# Local Scripts
```
steve-downey, line 1: We should have a list of things needed in order to run locally. I think it's
```
apt install clang-format-18 
apt install markdownlint
apt install shellcheck
apt install yamllint
apt install cmake-format
```
?

scripts/lint.sh
```diff
@@ -0,0 +1,215 @@
...
+    )
+    
+    FIX_INPLACE_FLAG=""
+    if [ "${FIX_INPLACE}" = true ]; then
+        FIX_INPLACE_FLAG="-i"
+    else
+        FIX_INPLACE_FLAG="--dry-run"
+    fi
+
+    find "${CPP_DIRS[@]}" -regex '.*\.\(hpp\|cpp\)$' | xargs clang-format --Werror -style=file --verbose ${FIX_INPLACE_FLAG}
```
steve-downey, line 105: I think it's `--style=` although I think it accepts the single dash?

scripts/README.md
neatudarius: Pushed that into `scripts/README.md`.

.markdownlint.json
neatudarius: Converted to `markdownlint.yml`, where comments are OK to be displayed. Also, pulled config from https://github.com/DavidAnson/markdownlint/blob/main/schema/.markdownlint.yaml. I'll tweak it.

etc/clang-17-toolchain.cmake
neatudarius: No.

If removed, `--fix` will do this:

```diff
-# etc/clang-17-toolchain.cmake -*- cmake -*-
-# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+# etc/clang-17-toolchain.cmake -*- cmake -*- SPDX-License-Identifier: Apache-2.0
+# WITH LLVM-exception
```

.clang-format
neatudarius: This change is not intentional. Will revert it.

.clang-format
neatudarius: > This will make it slightly harder to keep the .clang-format up to date because clang-format will want to rewrite it when you dump the config.

I'm OK to remove this file, but wouldn't be consistent for other linters? I think I would like to change it manually (using --dump-config into a temporary file). What do you think?

btw, I tried with clang 17/18/19, and I cannot get the current config. Would you like to update it with clang-format-19 in this PR? (The required changes will be applied in later commits or in a new PR).

steve-downey: It's the comment itself that gets lost. `clang-format` configs are generally one way upgradable but otherwise stable. That is, clang-format-N+1 will read an N file and default the elements that are added, so that when you run dump-config you get a file that only cf-N+1 understands but that produces the same output as the old config. 

Caveats around syntax that clang-format didn't used to understand, of course. 

We don't need to be eager about upgrading unless we decide we need the new knobs. 

I've left myself a TODO to see if clang-format changes matter for the few files we have, and if so, we probably don't need to upgrade. 

It's just that the comments are easy to get lost, and I wouldn't want anyone to read too much into that.

steve-downey:
I'll take a look at what the lint fixes would do before approving this, but I'm in principle in favor of all it. 

I have no strong feelings about markdown, shell, yaml, and json formatting -- all of them are terrible and I don't particularly care. 

Slight preferences about C++ formatting, although as long as clang-format can put things back in canonical form it's only a slight preference. And with `Lint Cpp` currently green, it looks like I'm OK.

scripts/lint.sh
```diff
@@ -0,0 +1,229 @@
...
+
+# scripts/lint.sh -*-shell-*-
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+# Lint all files in the project.
+# Expected to be run from the root of the project. 
+# Expected tools to be installed on the system. A pre-check is done to ensure all tools are installed.
+
+set -e
+set -x
```
steve-downey, line 11: Remove the execution trace option? 
It's very noisy to use the script with this set.

steve-downey (approved):
:+1:

neatudarius:
Related PR - #52 . Will discuss with @steve-downey the approach.

neatudarius:
Put this PR on hold until we solve similar issue/discussion for beman.exampler.

neatudarius:
> > Put this PR on hold until we solve similar issue/discussion for beman.exampler.
> 
> I think a good way to do that would be to mark it as a draft PR.

It seems it cannot be put back to draft. I will temporary close it.

PR #27, opened 2024-06-28
Enable clang-19 testing on CI

neatudarius:
Enable clang-19 testing on CI ( #1 )

It seems that the previous errors that we had with running clang-19 + Asan are connected to gtest. Check [gist](https://gist.github.com/neatudarius/87cbb37a1821e4dcbf773906fb15f8ae): 
```shell
#10 0x55f30b5bbd1e in __cxx_global_var_init /home/runner/work/Beman.Optional26/Beman.Optional26/src/Beman/Optional26/tests/optional.t.cpp:14:1
```
Literally here we just test gtest :( I also created an executable with a single file (outside the Beman.Optional26 tree) and got the same results.

Found the fix looking through some posts:
* https://github.com/llvm/llvm-project/issues/58404
* https://stackoverflow.com/questions/41552966/getting-new-delete-type-mismatch-from-asan


steve-downey (approved):
+1

Issue #28, opened 2024-06-28
Make Optional26 Packageable


bretbrownjr:
I attempted to add Optional26 to the Conan local recipe index I am finishing up in https://github.com/beman-project/beman/pull/21.

Two problems (so far):

- [ ] Optional26 needs tarball and/or zip releases.
- [ ] Optional26 uses CMakeLists.txt that assumes dependencies, at least `googletest` are vendored into the repo via git submodule workflows

The first issue can be fairly trivially resolved by clicking using the release feature in the github.com web UI and cutting a release. Of course, an established process and even CI for this task would be great, including developing patterns and CI scripts that other beman projects could benefit from.

The second issue could be a matter of making the relevant `add_subdirectory` conditional somehow, using `find_package` workflows instead and adding `googletest` as a package dependency in Conan to ensure it's available to build against. It's also possible that using `FetchContent` instead of git-submodules would make the packaging "work", but Conan is designed such that it does source code indexing and fetching, so I would prefer to adhere to that design when using Conan.

neatudarius:
Hello @bretbrownjr !

> The first issue can be fairly trivially resolved by clicking using the release feature in the github.com web UI and cutting a release. Of course, an established process and even CI for this task would be great, including developing patterns and CI scripts that other beman projects could benefit from.

I propose to discuss on Monday in the Beman sync. I think we should have an unified approach (including scripts, CI jobs) for releases.

> The second issue could be a matter of making the relevant add_subdirectory conditional somehow, using find_package workflows instead and adding googletest as a package dependency in Conan to ensure it's available to build against. It's also possible that using FetchContent instead of git-submodules would make the packaging "work", but Conan is designed such that it does source code indexing and fetching, so I would prefer to adhere to that design when using Conan.

I created  #29 29. Does this help as an intial step?

neatudarius:
@bretbrownjr , #29 was merged. I think we need to update the issue description.

gummif:
Any update on making this available in conan center?

PR #29, opened 2024-06-28
Split build production vs testing

neatudarius:
Split build production vs testing: #28 

- [x] Remove gtest submodule. Use Cmake `FetchContent_Declare` instead (gtest tag = $latest official release https://github.com/google/googletest/releases/tag/v1.14.0).
- [x] Use cmake `BUILD_TESTING` (default `ON` - build and run tests; can be set to `OFF`).
- [x] Remove redundant CMakeLists.txt as discussion in [discourse](https://discourse.bemanproject.org/t/cmake-add-subdirectory-shorthands/142)
- [x] Add build docs.


steve-downey:
With optional being a template, I'm concerned that not building tests will even more easily lead to silent breaks in building the library that only show up in users of the library. Although optional is pretty simple, I wouldn't want to encourage packagers to skip tests, even by accident. It's more important that a Conan or Vcpkg build self-test because it's even more likely that something is outside of our test matrix. 

I do have some code from an earlier version of my scratch project that will check if the gtest project exists before enabling the code, though. 

Might also switch, temporarily, to pulling googletest in as a `git subtree` rather than as a `git submodule`, which would mean we have a copy, but can also pull in updates, at least while we sort out a package reuse mechanism. Or switch to something lighter weight to vendor in.

neatudarius:
> With optional being a template, I'm concerned that not building tests will even more easily lead to silent breaks in building the library that only show up in users of the library. Although optional is pretty simple, I wouldn't want to encourage packagers to skip tests, even by accident. It's more important that a Conan or Vcpkg build self-test because it's even more likely that something is outside of our test matrix.

I'm OK with both approaches, but I would like to use an uniform approach for all libraries in Beman - CC: @camio , maybe we can add this topic to our sync agenda.

CMakeLists.txt
```diff
@@ -3,18 +3,30 @@
...
   VERSION 0.0.0
   LANGUAGES CXX)
 
-enable_testing()
+cmake_policy(VERSION 3.27)
 
 set(TARGETS_EXPORT_NAME ${CMAKE_PROJECT_NAME}Targets)
 
-add_subdirectory(extern)
+if("${CMAKE_BUILD_TESTS}")
```
JeffGarland, line 17: Are the quotes actually required here?

src/Beman/Optional26/tests/CMakeLists.txt
```diff
@@ -0,0 +1,22 @@
+# cmake-format: off
+# src/Beman/Optional26/tests/CMakeLists.txt -*-makefile-*-
```
camio, line 2: Can we remove the above line?

src/Beman/Optional26/tests/CMakeLists.txt
```diff
@@ -0,0 +1,22 @@
+# cmake-format: off
```
camio, line 1: Is there a cmake-format problem with the SPDX license identifier?

src/Beman/Optional26/tests/CMakeLists.txt
```diff
@@ -0,0 +1,22 @@
...
+# src/Beman/Optional26/tests/CMakeLists.txt -*-makefile-*-
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+# cmake-format: on
+
+# Tests
+add_executable(optional_test "")
```
camio, line 7: Is the `""` necessary? Also, did you consider adding all the sources as part of the `add_executable` call?

src/Beman/Optional26/tests/CMakeLists.txt
```diff
@@ -0,0 +1,22 @@
...
+
+target_sources(
+  optional_test
+  PRIVATE optional.t.cpp optional_ref.t.cpp
+          optional_monadic.t.cpp optional_range_support.t.cpp
+          optional_ref_monadic.t.cpp detail/iterator.t.cpp)
+
+target_link_libraries(optional_test "${TARGET_LIBRARY}" GTest::gtest GTest::gtest_main)
```
camio, line 15: Did you consider using `beman_optional26` instead of  using "${TARGET_LIBRARY}"? I've seen this pattern used in CMake code, but it would be more readable to use the variable with a good name (`beman_optional26`) instead of something generic and ambiguous like TARGET_LIBRARY.

src/Beman/Optional26/tests/CMakeLists.txt
```diff
@@ -0,0 +1,22 @@
...
+
+include(GoogleTest)
```
camio, line 17: Did you consider placing this first thing in the file? It's nice to see dependency listings in the prologue of a source code file.

src/Beman/Optional26/tests/CMakeLists.txt
```diff
@@ -0,0 +1,22 @@
...
+
+# Note: clang-19 + gtest_discover_tests + Asan setup causes errors on some platforms.
+# Temporary switch to gtest_add_tests and skip some Asan checks.
+enable_testing()
```
camio, line 21: This was already called in the top-level `CMakeLists.txt` so it doesn't seem necessary here.

src/Beman/Optional26/tests/CMakeLists.txt
```diff
@@ -0,0 +1,22 @@
...
+gtest_add_tests(optional_test "" AUTO)
```
camio, line 22: This is using the old form of `gtest_add_tests` which is maintained only for backwards compatibility. Did you consider using the new form?

src/Beman/Optional26/tests/CMakeLists.txt
```diff
@@ -0,0 +1,22 @@
...
+  PRIVATE optional.t.cpp optional_ref.t.cpp
+          optional_monadic.t.cpp optional_range_support.t.cpp
+          optional_ref_monadic.t.cpp detail/iterator.t.cpp)
+
+target_link_libraries(optional_test "${TARGET_LIBRARY}" GTest::gtest GTest::gtest_main)
+
+include(GoogleTest)
+
+# Note: clang-19 + gtest_discover_tests + Asan setup causes errors on some platforms.
+# Temporary switch to gtest_add_tests and skip some Asan checks.
```
camio, line 20: Did you consider creating a GitHub issue to track this and linking to it from here? That would allow for future maintainers to know when a workaround is no longer necessary.

Also, the comment mentions a temporary switch to gtest_add_tests, but it isn't clear what the preferred alternative is. Additionally I don't see any code below mentioning ASAN.

src/Beman/Optional26/tests/CMakeLists.txt
camio: `SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception` is the minimum necessary. Maybe we should just stick with that?

Before:

```CMake
# cmake-format: off
# src/Beman/Optional26/CMakeLists.txt -*-makefile-*-
# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
# cmake-format: on

code
```

After:

```CMake
# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception

code
```

neatudarius: If don't mind, I will actually configure clang-format and address this issue in https://github.com/beman-project/Optional26/pull/26. WIP issue.

src/Beman/Optional26/tests/CMakeLists.txt
neatudarius: I suppose this is a duplicated thread of https://github.com/beman-project/Optional26/pull/29/files#r1678269909. Please reopen if I am wrong.

src/Beman/Optional26/tests/CMakeLists.txt
neatudarius: > Did you consider creating a GitHub issue to track this and linking to it from here? That would allow for future maintainers to know when a workaround is no longer necessary.

Thanks for suggestion! Created #32 .

> so, the comment mentions a temporary switch to gtest_add_tests, but it isn't clear what the preferred alternative is. Additionally I don't see any code below mentioning ASAN.

Please check https://github.com/beman-project/Optional26/blob/main/.github/workflows/ci.yml#L78. We are running with ASAN enabled on CI.

src/Beman/Optional26/tests/CMakeLists.txt
neatudarius: I temporary used this framework because it caused some issues when running on CI - check #27 .

CMakeLists.txt
```diff
@@ -3,19 +3,35 @@
...
+# Include the CTest module.
+include(CTest)
+
+# Build the tests only if enabled via the CLI flag: BUILD_TESTING.
+if(BUILD_TESTING)
+  enable_testing()
+
+  # Fetch GoogleTest
+  include(FetchContent)
+  FetchContent_Declare(
```
steve-downey, line 24: I think without the EXCLUDE_FROM_ALL this is going to build and run the tests from googletest. 
But it also looks like that's not fixed until cmake 3.28.

steve-downey (approved):
Pulled the PR branch and confirmed it all works for me locally. 
Looks like gtest stopped including their tests in the `all` target they export, so we don't pick up all their infrastructure.

CMakeLists.txt
```diff
@@ -3,19 +3,33 @@
...
-enable_testing()
+cmake_policy(VERSION 3.27)
 
 set(TARGETS_EXPORT_NAME ${CMAKE_PROJECT_NAME}Targets)
 
-add_subdirectory(extern)
-add_subdirectory(src)
+# Includes
+include(CTest)
+include(FetchContent)
```
camio, line 19: I think these should go above the `TARGETS_EXPORT_NAME` variable definition.

src/Beman/Optional26/CMakeLists.txt
```diff
@@ -3,24 +3,22 @@
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 # cmake-format: on
 
-set(TARGET_LIBRARY "beman_optional26")
+add_library(beman_optional26 STATIC "")
 
-add_library("${TARGET_LIBRARY}" STATIC "")
-
-target_sources("${TARGET_LIBRARY}" PRIVATE optional.cpp detail/iterator.cpp)
+target_sources(beman_optional26 PRIVATE optional.cpp detail/iterator.cpp)
```
camio, line 8: Per https://github.com/beman-project/Optional26/pull/29#discussion_r1678271229, I think lines 6 and 8 can be merged.

src/Beman/Optional26/tests/CMakeLists.txt
camio: @steve-downey Making more complicated code now just in case we need it in the future doesn't have a good set of tradeoffs. Prioritizing readability encourages potential contributors to jump in.

src/Beman/Optional26/CMakeLists.txt
```diff
@@ -32,24 +30,9 @@ install(
   FILES_MATCHING
   PATTERN "*.hpp")
```
camio, line 31: Technically, we should be using FILE_SETs here. Not for this PR though.

src/Beman/Optional26/tests/CMakeLists.txt
camio: I don't mind if this is handled in another PR. I like the simplicity of my suggestion though.

src/Beman/Optional26/tests/CMakeLists.txt
camio: I think we still need the link in the comment above to the issue.

src/Beman/Optional26/tests/CMakeLists.txt
camio: Licensing info is usually at the top in Open Source projects. Bloomberg is the only exception I'm aware of.

src/Beman/Optional26/tests/CMakeLists.txt
camio: I'm suggesting you use `gtest_add_tests` with the `TARGET` keyword instead of the older form.

src/Beman/Optional26/tests/CMakeLists.txt
neatudarius: done

.github/workflows/ci.yml
```diff
@@ -12,12 +12,12 @@ jobs:
       fail-fast: false
       matrix:
         config:
-          - {name: "Ubuntu Clang 17", os: ubuntu-24.04, toolchain: "clang-17-toolchain.cmake", clang_version: 17, installed_clang_version: 14, cmake_args: "-G \"Ninja Multi-Config\" -DCMAKE_CONFIGURATION_TYPES=\"RelWithDebInfo;Asan\" "}
-          - {name: "Ubuntu Clang 18", os: ubuntu-24.04, toolchain: "clang-18-toolchain.cmake", clang_version: 18, installed_clang_version: 14, cmake_args: "-G \"Ninja Multi-Config\" -DCMAKE_CONFIGURATION_TYPES=\"RelWithDebInfo;Asan\" "}
+          - {name: "Ubuntu Clang 17", os: ubuntu-24.04, toolchain: "clang-17-toolchain.cmake", clang_version: 17, installed_clang_version: 17, cmake_args: "-G \"Ninja Multi-Config\" -DCMAKE_CONFIGURATION_TYPES=\"RelWithDebInfo;Asan\" "}
```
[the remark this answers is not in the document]
neatudarius: Agree. We'll solve this issue in a new PR.

CMakeLists.txt
neatudarius: yes, it works.

README.md
```diff
@@ -178,6 +198,63 @@ make -k TOOLCHAIN=clang-18  CONFIG=Tsan  VERBOSE=1
...
 
+#### Step by Step Build: Build and Run Tests
+
+CI current build and test flows:
+
+```shell
+# Configure build: default build production code + tests (BUILD_TESTING=ON by default).
+$ rm -rf .build/
+$ mkdir -p .build
+$ cd .build
```
camio, line 209: You can avoid changing the working directory to `.build` and back using CMake's `-B` parameter.

README.md
```diff
@@ -178,6 +198,63 @@ make -k TOOLCHAIN=clang-18  CONFIG=Tsan  VERBOSE=1
...
 The makefile will use your system compiler, `c++`, if no toolchain name is provided, otherwise it will use the toolchain in the etc/ directory to perform the build. The Ninja multi config generator is used, with configurations for `RelWithDebugInfo`, `Debug`, `Tsan`, and `Asan` configured by default.
 
+#### Step by Step Build: Build and Run Tests
+
+CI current build and test flows:
+
+```shell
+# Configure build: default build production code + tests (BUILD_TESTING=ON by default).
+$ rm -rf .build/
+$ mkdir -p .build
```
camio, line 208: I don't think this command is necessary. cmake will create the directory IIUC.

src/Beman/Optional26/tests/CMakeLists.txt
```diff
@@ -0,0 +1,30 @@
+# cmake-format: off
+# src/Beman/Optional26/tests/CMakeLists.txt -*-makefile-*-
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+# cmake-format: on
+
+include(GoogleTest)
+
+# Tests
+add_executable(optional_test
```
camio, line 9: The target should probably be called beman_optional26_test or something like that to avoid collisions in case this project is embedded in a larger project. Target names are global.

camio (approved):
I mentioned a couple minor things, but it otherwise looks good. Note that I didn't review the docker file at all.

README.md
steve-downey: Although starting deleting the build directory can fix a lot of strange problems, as CMake will only check some things the very first time its run in directory. `realclean` and not just the "clean" target can be necessary if you're changing anything significant. 
If I'm building with different toolchains I build in ./build/$toolchain, or ../build/$project/$toolchain, so as not to have crosstalk between incompatible builds.

src/Beman/Optional26/tests/CMakeLists.txt
steve-downey: We may want to use PROJECT_IS_TOP_LEVEL to gate that in order to be friendlier with Fetch. For packaging tests ought to be run, but for vendoring that needs to be more controllable. 

https://cmake.org/cmake/help/latest/variable/PROJECT_IS_TOP_LEVEL.html

neatudarius:
> I mentioned a couple minor things, but it otherwise looks good. Note that I didn't review the docker file at all.

Sorry, I added the Dockerfile by mistake. I don't want to added right now. Currently I'm using it to replicate the environment from CI, but it's not ready. TBD

src/Beman/Optional26/tests/CMakeLists.txt
neatudarius: > The target should probably be called beman_optional26_test or something like that to avoid collisions in case this project is embedded in a larger project. Target names are global.

Done.

> We may want to use PROJECT_IS_TOP_LEVEL to gate that in order to be friendlier with Fetch. For packaging tests ought to be run, but for vendoring that needs to be more controllable.

I'm not sure what would be the improvement here. We already have conditional in parent CmakeLists.txt:
```shell
# in Beman/Beman.Optional26/src/Beman/Optional26/CMakeLists.txt

# Tests
if(BUILD_TESTING)
  add_subdirectory(tests)
endif()
```
@steve-downey , am I missing something?

README.md
neatudarius: Indeed, simplified to:

CI current build and test flows:

```shell
# Configure build: default build production code + tests (BUILD_TESTING=ON by default).
$ cmake -G "Ninja Multi-Config" -DCMAKE_CONFIGURATION_TYPES="RelWithDebInfo;Asan" -DCMAKE_TOOLCHAIN_FILE=etc/clang-19-toolchain.cmake -B .build -S .
-- The CXX compiler identification is Clang 19.0.0
...
-- Build files have been written to: /path/to/Optional26/.build

# Build.
$ cmake --build .build --config Asan --target all -- -k 0
...
[30/30] Linking CXX executable ...

# Run tests.
$ ctest --build-config Asan --output-on-failure --test-dir .build
Internal ctest changing into directory: /path/to/Optional26/.build
Test project /path/to/Optional26/.build
...
100% tests passed, 0 tests failed out of 82

Total Test time (real) =   0.67 sec
```

Thanks!

README.md
neatudarius: simplified with -B

Issue #45, opened 2024-07-18
is_optional would be more efficient as a variable template


jwakely:
https://github.com/beman-project/Optional26/blob/bfeb9080c0d9c501965ae857a1decf7ed74686f5/include/Beman/Optional26/optional.hpp#L237-L242

There seems to be no reason to prefer a class template to a variable template, it just makes it slower to compile. Why `decay_t` not `remove_cvref_t`? Or why do that at all? All uses of `is_optional` already use either `remove_cvref` or `decay` (which should be `remove_cvref` in all cases), except for the use in `optional<T&>::and_then` which is probably incorrect.

```c++
template <class T>
inline constexpr bool is_optional = false;
template <class T>
inline constexpr bool is_optional<optional<T>> = true;
```

steve-downey:
Testing the `static_assert`s moved up the priority list, too. It was ad-hoc, and therefore wrong.

Issue #44, opened 2024-07-18
in_place tag needs to be inline not static


jwakely:
https://github.com/beman-project/Optional26/blob/bfeb9080c0d9c501965ae857a1decf7ed74686f5/include/Beman/Optional26/optional.hpp#L200

This will cause ODR violations and problems with modules.

wusatosi:
Look like this should be closed as its already completed

Issue #43, opened 2024-07-18
Unnecessary default constructor overloading


jwakely:
https://github.com/beman-project/Optional26/blob/bfeb9080c0d9c501965ae857a1decf7ed74686f5/include/Beman/Optional26/optional.hpp#L325-L331

Is there any benefit to the defaulted constructor? Why define it twice? The default member-initializers mean it's never trivial, even if defined as defaulted.

Unless I'm missing something, it looks like this could be simply:

`constexpr optional() noexcept {}`

steve-downey:
I don't think there's a requirement for optional to be trivial, so I don't think that's it. It may have been an older compiler being unhappy. Will change and confirm in tests.

Issue #42, opened 2024-07-18
or_else(F&&) && should be constexpr


jwakely:
https://github.com/beman-project/Optional26/blob/bfeb9080c0d9c501965ae857a1decf7ed74686f5/include/Beman/Optional26/optional.hpp#L580

PR #47, opened 2024-07-22
Fix issues

steve-downey:


steve-downey:
Fixes in the primary template for issues #36 #37 #38 #29 #40 #41 #42 #43 #44

Plan to add fix for #45 also. 

Need to review the reference specialization for similar issues.

.github/workflows/ci.yml
```diff
@@ -16,6 +16,8 @@ jobs:
...
           # Note: clang-19 + Asan setup causes errors on some platforms. Temporary skip some checks via .asan_options.
           - {name: "Ubuntu Clang 19", os: ubuntu-24.04, toolchain: "clang-19", clang_version: 19, installed_clang_version: 17, cmake_args: "-G \"Ninja Multi-Config\" -DCMAKE_CONFIGURATION_TYPES=\"RelWithDebInfo;Asan\" ", asan_options: "new_delete_type_mismatch=0"}
+          - {name: "Ubuntu GCC 11", os: ubuntu-24.04, toolchain: "gcc-11",  cmake_args: "-G \"Ninja Multi-Config\" -DCMAKE_CONFIGURATION_TYPES=\"RelWithDebInfo;Asan\" "}
+          - {name: "Ubuntu GCC 12", os: ubuntu-24.04, toolchain: "gcc-12",  cmake_args: "-G \"Ninja Multi-Config\" -DCMAKE_CONFIGURATION_TYPES=\"RelWithDebInfo;Asan\" "}
           - {name: "Ubuntu GCC 13", os: ubuntu-24.04, toolchain: "gcc-13",  cmake_args: "-G \"Ninja Multi-Config\" -DCMAKE_CONFIGURATION_TYPES=\"RelWithDebInfo;Asan\" "}
           - {name: "Ubuntu GCC 14", os: ubuntu-24.04, toolchain: "gcc-14", cmake_args: "-G \"Ninja Multi-Config\" -DCMAKE_CONFIGURATION_TYPES=\"RelWithDebInfo;Asan\" "}
```
neatudarius, line 22: should we also add gcc-15 here?

include/Beman/Optional26/optional.hpp
```diff
@@ -546,38 +538,56 @@ class optional {
     template <class F>
     constexpr auto transform(F&& f) & {
         using U = std::invoke_result_t<F, T&>;
+        static_assert(!std::is_array_v<U>);
+        static_assert(!std::is_same_v<U, in_place_t>);
+        static_assert(!std::is_same_v<U, nullopt_t>);
+        //        static_assert(std::is_object_v<U>); /// References now allowed
```
neatudarius, line 544: reminder

.github/workflows/ci.yml
neatudarius, line 22: should we also add gcc-15 here?

src/Beman/Optional26/tests/optional_constexpr.t.cpp
```diff
@@ -0,0 +1,1744 @@
...
+TEST(OptionalConstexprTest, Constructors) {
+    constexpr beman::optional26::optional<int> i1;
+    constexpr beman::optional26::optional<int> i2{beman::optional26::nullopt};
+
+    constexpr int                              i  = 0;
+    constexpr beman::optional26::optional<int> i3 = i;
+    (void)i3;
+
+    constexpr beman::optional26::optional<empty> e1;
+    constexpr beman::optional26::optional<int>   e2{beman::optional26::nullopt};
```
neatudarius, line 47:
```suggestion
    constexpr beman::optional26::optional<empty>   e2{beman::optional26::nullopt};
```
?

src/Beman/Optional26/tests/optional_constexpr.t.cpp
```diff
@@ -0,0 +1,1744 @@
...
+#include <functional>
+#include <ranges>
+#include <tuple>
+
+#include <gtest/gtest.h>
+
+TEST(OptionalConstexprTest, TestGTest) { ASSERT_EQ(1, 1); }
+
+namespace {
+struct empty {};
```
neatudarius, line 15: Please check if you can use / extend types from `Beman/Beman.Optional26/src/Beman/Optional26/tests/test_types.hpp`.

I know this file is a copy of `optional.t.cpp`, but maybe we should do the updates for both files.

src/Beman/Optional26/tests/optional_constexpr.t.cpp
```diff
@@ -0,0 +1,1744 @@
...
+};
+
+struct derived : public base {
+    int j_;
+    constexpr derived() : base(0), j_(0) {}
+    constexpr derived(int i, int j) : base(i), j_(j) {}
+};
+} // namespace
+
+TEST(OptionalConstexprTest, Constructors) {
```
neatudarius, line 38: Can we reduce the duplication between optional.t.cpp and optional_constexpr.t.cpp?

src/Beman/Optional26/tests/optional_constexpr.t.cpp
```diff
@@ -0,0 +1,1744 @@
...
+    constexpr beman::optional26::optional<int> i1;
+    constexpr beman::optional26::optional<int> i2{beman::optional26::nullopt};
+
+    constexpr int                              i  = 0;
+    constexpr beman::optional26::optional<int> i3 = i;
+    (void)i3;
```
neatudarius, line 44: Should we use std::ignore ?

src/Beman/Optional26/tests/optional_constexpr.t.cpp
```diff
@@ -0,0 +1,1744 @@
...
+};
+} // namespace
+
+TEST(OptionalConstexprTest, NonDefaultConstruct) {
+    constexpr NoDefault                              i = 7;
+    constexpr beman::optional26::optional<NoDefault> v1{};
+    constexpr beman::optional26::optional<NoDefault> v2{i};
+}
+
+TEST(OptionalConstexprTest, Triviality) {
```
neatudarius, line 133: Can we reduce the duplication between optional.t.cpp and optional_constexpr.t.cpp?


Check https://gist.githubusercontent.com/neatudarius/1ded802e47996445ed4a731d10f7c4fc/raw/36448222892a0fe403ea39c5f0eb2c0dc89e9097/gistfile1.txt, e.g. `TEST(OptionalConstexprTest, Triviality) ` vs `TEST(OptionalTest, Triviality) `

steve-downey: Factoring out repeated nonce types is something we should definitely do. Testing constexpr  is tricky, though. You have to call it in a consteval context, and that doesn't necessarily tell you if the runtime evaluation is correct, because they can be effectively unrelated. 
X macros or preprocessor tricks? I don't have a good idea off-hand for the parallelism.

src/Beman/Optional26/tests/optional_constexpr.t.cpp
steve-downey: I'll check if it's available and working. Nicer than the (void) cast hack.

.github/workflows/ci.yml
steve-downey: I'm not sure where to get a good build for it? I build and install mine locally. 
The compiler testing PPA might work in this case. It can seriously destabilize a system by interfering with the production compiler and the next update, but this is a container that's being thrown away.

src/Beman/Optional26/tests/optional_constexpr.t.cpp
steve-downey: They diverge slightly, but there are a lot of Don't Repeat Yourself violations. 
Refactor as part of this commit, or follow on?

src/Beman/Optional26/tests/optional_constexpr.t.cpp
steve-downey: Probably want all of them. Not that the test is really interesting, but empty type constructed into an empty optional was intentional. I think.

include/Beman/Optional26/optional.hpp
steve-downey: It's the comment on the end that's important. This has to make it back into the optional<T&> paper as it's a change I hadn't realized needed to be made to the base template.

Although I've been thinking about it and it might need to be more complicated. It needs to be, I think now, either a reference -- to any type -- or an object that can be stored in the optional. 
std::is_reference_v<U> || std::is_object_v<U>

At least disjunction in a static_assert isn't expensive like in concept constraints.

src/Beman/Optional26/tests/optional_constexpr.t.cpp
```diff
@@ -0,0 +1,1744 @@
...
+    constexpr beman::optional26::optional<int> o1{4};
+    constexpr beman::optional26::optional<int> o2{42};
+    constexpr beman::optional26::optional<int> o3{};
+
+    //  SECTION("self simple")
+    {
+        {
+        constexpr auto b = !(o1 == o2);
+        EXPECT_TRUE(b);
+        }
```
steve-downey, line 349:
```C++
consteval auto constantify(auto expr){return expr;}
/// ...
EXPECT_TRUE(constantify(!(o1 == o2))
```
might be more readable, and a little less noise. At least less vertical space consumption for all these tests.

src/Beman/Optional26/tests/test_types.hpp
```diff
@@ -4,7 +4,9 @@
 #ifndef BEMAN_OPTIONAL26_TESTS_TEST_TYPES_HPP
 #define BEMAN_OPTIONAL26_TESTS_TEST_TYPES_HPP
 
-namespace beman::optional26::test {
+#include <compare>
+
+namespace beman::optional26::tests {
```
neatudarius, line 9: why longer name? :D

src/Beman/Optional26/tests/test_types.cpp
```diff
@@ -0,0 +1 @@
+#include <Beman/Optional26/tests/test_types.hpp>
```
neatudarius, line 1: Please add cpp header (2 lines).

src/Beman/Optional26/tests/optional_constexpr.t.cpp
neatudarius: Ok, we can defer this task.

src/Beman/Optional26/tests/optional_constexpr.t.cpp
neatudarius: As you prefer.

include/Beman/Optional26/optional.hpp
neatudarius: Then add a TODO inside the comment.

src/Beman/Optional26/tests/optional_ref.t.cpp
```diff
@@ -2,68 +2,53 @@
...
-    derived(int i, int j) : base(i), j_(j) {}
-};
-} // namespace
 
 TEST(OptionalRefTest, Constructors) {
     beman::optional26::optional<int&> i1;
     beman::optional26::optional<int&> i2{beman::optional26::nullopt};
-    (void)i1;
-    (void)i2;
+    std::ignore = i1;
```
neatudarius, line 15: Thanks for adding this!

src/Beman/Optional26/tests/optional_ref.t.cpp
```diff
@@ -568,3 +553,106 @@ TEST(OptionalRefTest, SwapNullIntializedWithValue) {
...
+    } else {
+        optional_base_ref.reset();
+    }
+    EXPECT_TRUE(optional_base_ref.has_value());
+
+    derived d2(2, 2);
+    engaged_derived = d2;
+    EXPECT_EQ(optional_base_ref.value().m_i, static_cast<base>(d2).m_i);
+
+    // delete the rvalue ref overload
```
neatudarius, line 617: Please add TODOs where not implemented yet to distinguish from dead code.

src/Beman/Optional26/tests/optional_ref.t.cpp
```diff
@@ -568,3 +553,106 @@ TEST(OptionalRefTest, SwapNullIntializedWithValue) {
...
+    beman::optional26::optional<derived> engaged_derived{d};
+
+    static_assert(std::is_constructible_v<const base&, derived>);
+
+    beman::optional26::optional<const base&> optional_base_const_ref{empty_derived};
+    EXPECT_FALSE(optional_base_const_ref.has_value());
+
+    beman::optional26::optional<const base&> optional_base_const_ref2{engaged_derived};
+    EXPECT_TRUE(optional_base_const_ref2.has_value());
+
```
neatudarius, line 657:
```suggestion
```

src/Beman/Optional26/tests/optional.t.cpp
```diff
@@ -854,338 +822,3 @@ TEST(ViewMaybeTest, RefBase) {
     }
     ASSERT_EQ(s, 9);
 }
-
-// TEST(ViewMaybeTest, MonadicAndThen) {
```
neatudarius, line 858: Thanks for split!

src/Beman/Optional26/tests/optional_constexpr.t.cpp
```diff
@@ -0,0 +1,1261 @@
...
+#include <Beman/Optional26/optional.hpp>
+#include <Beman/Optional26/tests/test_types.hpp>
+
+#include <functional>
+#include <ranges>
+#include <tuple>
+
+#include <gtest/gtest.h>
+
+TEST(OptionalConstexprTest, TestGTest) { ASSERT_EQ(1, 1); }
```
neatudarius, line 13: I don't know if we need to add this test in each file. Do you want to propose this as a convention?

src/Beman/Optional26/tests/optional_constexpr.t.cpp
```diff
@@ -0,0 +1,1261 @@
...
+        constexpr not_trivial_copy_assignable(int i) : i_(i) {}
+        constexpr not_trivial_copy_assignable(const not_trivial_copy_assignable&) = default;
+        constexpr not_trivial_copy_assignable& operator=(const not_trivial_copy_assignable& rhs) {
+            i_ = rhs.i_;
+            return *this;
+        }
+    };
+
+    /*
+      optional& operator=(const optional& rhs)
```
neatudarius, line 178: What do you think to always use `//` multi line comment?

src/Beman/Optional26/tests/optional_constexpr.t.cpp
```diff
@@ -0,0 +1,1261 @@
...
+    retval &= (std::get<1>(*o4) == 1);
+
+    constexpr beman::optional26::optional<takes_init_and_variadic> o6(beman::optional26::in_place, {0, 1}, 2, 3);
+    retval &= (o6->v0 == 0);
+    retval &= (std::get<0>(o6->t) == 2);
+    retval &= (std::get<1>(o6->t) == 3);
+    return retval;
+}
+
+auto consteval constify(auto expr) { return (expr); }
```
neatudarius, line 223: Please add docs for utility functions. Any reason not to move this in the test_type header?

src/Beman/Optional26/tests/optional_constexpr.t.cpp
```diff
@@ -0,0 +1,1261 @@
...
+
+    EXPECT_TRUE(constify(!o1));
+    EXPECT_TRUE(constify(!o2));
+    EXPECT_TRUE(constify(!o3));
+    EXPECT_TRUE(constify(!o4));
+
+    EXPECT_TRUE(!std::is_default_constructible<beman::optional26::nullopt_t>::value);
+}
+
+struct move_detector {
```
neatudarius, line 271: Same here : why is not int `test_types.h`?

src/Beman/Optional26/tests/optional_constexpr.t.cpp
```diff
@@ -0,0 +1,1261 @@
...
+    EXPECT_TRUE(success);
+    constexpr auto success2 = std::is_same<decltype(o3.value()), const int&>::value;
+    EXPECT_TRUE(success2);
+    constexpr auto success3 = std::is_same<decltype(std::move(o1).value()), const int&>::value;
+    EXPECT_TRUE(success3);
+
+}
+
+namespace {
+class Point
```
neatudarius, line 296: same for this. we may reuse the Point type in other `*.t.cpp` files.

src/Beman/Optional26/tests/optional_constexpr.t.cpp
```diff
@@ -0,0 +1,1261 @@
...
+            static_assert(b);
+        }
+    }
+    return true;
+}
+
+constexpr bool checkTestComparison = testComparisons();
+static_assert(checkTestComparison);
+
+TEST(OptionalConstexprTest, RangeTest) {
```
neatudarius, line 1118: Should this test be added at the end of optional_range_support.t.cpp file?

Also, I was expecting do have a constexpr function which creates 2 ranges, iterates over them and compute something, and check the result at compile time. is this mixt test useful/complete?

neatudarius (commented):
LGTM overall, check my current feedback:

src/Beman/Optional26/tests/optional_constexpr.t.cpp
steve-downey: It's actually a hold-over from the template I use where it's actually ASSERT_EQ(0,1) so I don't spend a lot of time working on tests that aren't being built or run. 

Deleting it would be fine once it's served its purpose. 

Adding a note to work out a lint check, on my long list.

src/Beman/Optional26/tests/optional_constexpr.t.cpp
steve-downey: That's fine with me.

src/Beman/Optional26/tests/optional_ref.t.cpp
steve-downey: ✔️

src/Beman/Optional26/tests/test_types.hpp
steve-downey: Matching the namespaces to the path name and include guard. It's not really important here, but an internal linter checks that canonical names of headers matches the namespaces used so that collisions and ambiguities are avoided.

src/Beman/Optional26/tests/optional_constexpr.t.cpp
neatudarius: Ok. let's keep it as it is.

src/Beman/Optional26/tests/test_types.hpp
neatudarius: cool. let's keep it

neatudarius (approved):
LGTM 👍

PR #53, opened 2024-08-23
Update optional<T&> paper with new Wording based on Beman Optional26

steve-downey:
Update the paper for optional<T&> based on the updates to Optional26. Reorganize optional26 so that the header and class synopsis in the standard can be quoted mostly unmodified from the tested code. This involves moving the definitions of functions out of line from the class declaration and pulling std:: functions into scope via using declarations, as well as making sure non standard proposed helpers are in a detail namespace. In the paper itself, the synopsis and definitions are slightly modified to indicate exposition only members and types, changing some explicit and throw specifications to see-below, and moving the expression to the function definition, and, since <optional> is quite old at this point, rendering requires clauses in declarations as constraint clauses in text. 

Also add two new "top level" papers that only produce the before and after changes for the wording, in order to simplify cross checking differences for review, and to later produce a better change request for the draft standard, in addition to the edit instructions in the paper. 

This also adds a partial specialization for optional<T&&> which is nearly identical to optional<T&> and would be even more so if all optional specializations  used optional::value_type throughout its definition where that is the actual intent.


neatudarius:
clang-format added by mistake?

src/Beman/Optional26/tests/optional_refref.t.cpp
```diff
@@ -0,0 +1,618 @@
...
+        EXPECT_TRUE(!("hello" > o4));
+        EXPECT_TRUE(o4 <= "hello");
+        EXPECT_TRUE("hello" <= o4);
+        EXPECT_TRUE(o4 >= "hello");
+        EXPECT_TRUE("hello" >= o4);
+    }
+}
+
+TEST(OptionalRefRefTest, Triviality) {
+    EXPECT_TRUE(std::is_trivially_copy_constructible<beman::optional26::optional<int&&>>::value);
```
neatudarius, line 247: Maybe use an alias `using Type = beman::optional26::optional<int&&>;` and then `EXPECT_TRUE(std::is_trivially_copy_constructible<Type>::value`);

Btw, we can put a follow-up commit/PR to remove duplication across `*.t.cpp`. We can create a new issue. for that.

include/Beman/Optional26/optional.hpp
```diff
@@ -229,719 +125,979 @@ inline constexpr auto format_kind<beman::optional26::optional<T>> = range_format
...
-        if (has_value())
-            value_.~T();
-    }
-
-    constexpr ~optional()
-        requires std::is_trivially_destructible_v<T>
-    = default;
-
-    constexpr optional(nullopt_t) noexcept {}
+    using iterator       = detail::contiguous_iterator<T, optional>; // see~\ref{optional.iterators}
```
neatudarius, line 288:
```suggestion
    using iterator       = detail::contiguous_iterator<T, optional>; // ~\ref{optional.iterators}
```
?

include/Beman/Optional26/optional.hpp
```diff
@@ -229,719 +125,979 @@ inline constexpr auto format_kind<beman::optional26::optional<T>> = range_format
...
-
-            std::construct_at(std::addressof(value_), rhs.value_);
-            engaged_ = true;
-        }
-    }
-
+        requires is_copy_constructible_v<T> && (!is_trivially_copy_constructible_v<T>);
     constexpr optional(const optional&)
-        requires std::is_copy_constructible_v<T> && std::is_trivially_copy_constructible_v<T>
+        requires is_copy_constructible_v<T> && is_trivially_copy_constructible_v<T>
```
neatudarius, line 298: why do we prefer this convention ? using multiple using std::something at the top of the file

include/Beman/Optional26/optional.hpp
```diff
@@ -950,209 +1106,466 @@ auto optional_map_impl(Opt&& opt, F&& f)
 template <class T>
 class optional<T&> {
```
neatudarius, line 1107: What do you think to split this file? Maybe:

* move `class optional<T>` into `details/optional_impl.hpp`
* move`class optional<T&>` into `details/optional_ref_impl.hpp`
* include both in `optional.hpp`

If we agree with such approach, I'm OK to defer the work into a new issue.

include/Beman/Optional26/optional.hpp
```diff
@@ -950,209 +1106,466 @@ auto optional_map_impl(Opt&& opt, F&& f)
...
+constexpr optional<T&>::optional() noexcept : value_(nullptr) {}
+
+template <class T>
+constexpr optional<T&>::optional(nullopt_t) noexcept : value_(nullptr) {}
+
+template <class T>
+template <class U>
+    requires(!detail::is_optional<decay_t<U>>)
+constexpr optional<T&>::optional(U&& u) noexcept : value_(addressof(u)) {
+    static_assert(is_constructible_v<add_lvalue_reference_t<T>, U>, "Must be able to bind U to T&");
```
frederick-vs-ja, line 1194: Why are `add_lvalue_reference_t` and `add_rvalue_reference_t` used given `T&` must be well-formed in these partial specializations?

steve-downey:
> clang-format added by mistake?

I'll double check, but I think this one is probably dealing with formatting for papers and the standard, which has some tradeoffs made decades ago, and has to deal with narrow margins.

neatudarius (approved):
LGTM 👍 

I'm going in PTO for a week. If not merged until I come back, I will do one more round of review. But feel free to proceed from my side.

steve-downey:
Superseded by #70

PR #54, opened 2024-08-27
Avoid reopening namespace `std`

frederick-vs-ja:
Following the suggestion in [_Don’t reopen `namespace std`_](https://quuxplusone.github.io/blog/2021/10/27/dont-reopen-namespace-std/).

The change also works around a bug of MSVC (reduced and [reported on DevCom](https://developercommunity.visualstudio.com/t/MSVC-doesnt-accept-specialization-defin/10732088)).

Godbolt links:
- Before: https://godbolt.org/z/1bf1458Pf
- After: https://godbolt.org/z/1eanosrob

Drive-by: Remove emoji `🔗` from the synopsis comments, which seems to be copied from https://eel.is/c++draft/optional.syn.

### DIFF
--- a/include/Beman/Optional26/optional.hpp
+++ b/include/Beman/Optional26/optional.hpp
@@ -6,7 +6,6 @@
 
 /*
 22.5.2 Header <optional> synopsis[optional.syn]
-🔗
 
 #include <compare>              // see [compare.syn]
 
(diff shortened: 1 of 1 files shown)


neatudarius (approved):
LGTM 👍 

Indeed, it does not compile with MSVC. I isolated a minimal code here https://godbolt.org/z/zvq5Mf4eW. 

CC: @steve-downey

steve-downey:
It looks good to me,  too. 
In a different PR delete the whole quoted synopsis, as it's out of date, and becomes redundant as the new code matches, but cleaning the stray paperclip is good now.

neatudarius:
@frederick-vs-ja , please proceed with the PR merge. I would also like to check if you have permissions this way.

frederick-vs-ja:
> @frederick-vs-ja , please proceed with the PR merge. I would also like to check if you have permissions this way.

I don't have write permission currently. Would you like to merge this?

PR #58, opened 2024-09-16
Fix some bugs in optional<T&>; add tests for them

Quuxplusone:
There's a lot of commits in this PR — feel free to cherry-pick (but I think they all need to be taken/addressed, one way or another). The really important ones are:

```
commit 4448d97dbd98e017203931e4a9f9a5a8ddb831e8
Author: Arthur O'Dwyer <arthur.j.odwyer@gmail.com>
Date:   Mon Sep 16 13:06:42 2024 -0400

    [test] Add a green test for overload resolution

commit 986b975ad718613351aa1fe4ce20d75e6f956c41
Author: Arthur O'Dwyer <arthur.j.odwyer@gmail.com>
Date:   Mon Sep 16 12:20:19 2024 -0400

    Support optional<T&>::emplace, and add tests
    
    The return type of `emplace` had been wrong; `emplace` returns a
    reference to the emplaced value (in this case, the T& reference
    that was bound), not `*this`.
    
    The new tests are red before the patch and green afterward.

commit a6333d30c8a09c38dae8577f67181a74dbb34e4b
Author: Arthur O'Dwyer <arthur.j.odwyer@gmail.com>
Date:   Mon Sep 16 12:11:47 2024 -0400

    Support binding optional<const T&> to a non-const T, and add tests
    
    The new tests are red before the patch and green afterward.

commit cd93469568a509ac3d2d18904df95e981b196136
Author: Arthur O'Dwyer <arthur.j.odwyer@gmail.com>
Date:   Mon Sep 16 11:29:36 2024 -0400

    Support optional<optional<int>&>
    
    Add two new tests: one for `optional<optional<int>&>`, and one
    testing that we can construct and assign to `optional<T&>` from
    `reference_wrapper<T>` (which should work: since `reference_wrapper<T>`
    is implicitly convertible to `T&`, it also should be implicitly
    convertible to `optional<T&>`).
    These tests are both red before this patch and green afterward.
```

These also correspond to bugs and/or typos in P2988's wording; for example it has `emplace` returning `optional&` instead of `T&`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,5 +1,6 @@
 name: CI Tests
 on:
+  workflow_dispatch:
   push:
     branches: [main]
   pull_request:
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -1,6 +1,7 @@
 name: pre-commit
 
 on:
+  workflow_dispatch:
   pull_request:
   push:
     branches: [main]
--- a/include/Beman/Optional26/optional.hpp
+++ b/include/Beman/Optional26/optional.hpp
@@ -971,6 +971,9 @@ template <class To, class From>
 concept safely_constructible_not_convertible = std::is_constructible_v<To, From> && !std::is_convertible_v<From, To> &&
                                                !reference_constructs_from_temporary_v<To, From>;
 
+template <class To, class From>
+concept safely_constructible = std::is_constructible_v<To, From> && !reference_constructs_from_temporary_v<To, From>;
+
 } // namespace detail
 
 template <class T>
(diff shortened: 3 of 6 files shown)


steve-downey:
I'll merge main back to my working branch and then merge this change set on
top of it.

steve-downey:
I've merged main and the additional tests. The changes to optional.hpp need to be done piece by piece as  the diffs are large enough to confuse git.

camio:
Thanks for the contributions @Quuxplusone, and welcome!

camio:
@steve-downey/@neatudarius, what is the status of this PR? Were these changes applied in a different merge?

steve-downey:
They are incorporated into https://github.com/steve-downey/Optional26/pull/19 
Which will be a PR into https://github.com/beman-project/optional26/ soon.

Quuxplusone:
@camio Also, FWIW, I believe we've established that my new test `OverloadResolutionChecksDangling` is _supposed_ to be red — that is, _I_ think it should go green, but @steve-downey and Tomasz say it was designed not-to-work on purpose (for consistency with how it doesn't-work for `tuple<T&>` either). That, in turn, means that my new helpers `safely_convertible` and `safely_constructible` are kind of pointless. So in short I think Steve and I are on the same page that this PR isn't _trivially_ mergeable, and I'm still unclear as to which parts of it _are_ uncontroversial.

E.g. I _think_ it's uncontroversial that the return type of `emplace` should be `T&`, but I'm not even 100% sure we're on the same page about that, necessarily.

PR #59, opened 2024-09-16
Prevent calling make_optional<int&>(i) or make_optional<const int&>(i)

Quuxplusone:
The status quo in C++23 is that

        int i;
        auto o = make_optional<int&>(i);

produces an `optional<int>`, not an `optional<int&>`. This would be horribly confusing to propagate into C++26 — users would naturally expect to receive an `optional<int&>`. Worse,

        auto o = make_optional<int&>(std::ref(i));

actually *would* give you an `optional<int&>`, because it would use a different overload of `make_optional`! So, we propose to add a "Mandates" element to `make_optional` that requires its explicit template argument (if any) to be a non-array object type.
This breaks the "status quo" example at the top of this comment (requiring an Annex C entry); but it successfully prevents the pitfalls. And it does not _silently change the behavior_ of the "status quo" example, either — which would probably be 100x worse than merely breaking it!

Thanks to Tomasz Kamiński for the template-programming trick, which matches the technique used in https://eel.is/c++draft/out.ptr .

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,5 +1,6 @@
 name: CI Tests
 on:
+  workflow_dispatch:
   push:
     branches: [main]
   pull_request:
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -1,6 +1,7 @@
 name: pre-commit
 
 on:
+  workflow_dispatch:
   pull_request:
   push:
     branches: [main]
--- a/include/Beman/Optional26/optional.hpp
+++ b/include/Beman/Optional26/optional.hpp
@@ -234,28 +234,39 @@ template <class T>
 inline constexpr bool is_optional<optional<T>> = true;
 } // namespace detail
 
-template <typename T>
-constexpr optional<std::decay_t<T>>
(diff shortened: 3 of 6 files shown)


frederick-vs-ja:
How about making the first overload like this, which seems simpler to me?
```C++
    namespace detail {
        struct spec_barrier {
            explicit spec_barrier() = default;
        };
    }

    template<detail::spec_barrier = detail::spec_barrier{}, class T>
        requires std::is_constructible_v<std::decay_t<T>, T> // per LWG3627
    constexpr std::optional<std::decay_t<T>> make_optional(T&& t)
        noexcept(std::is_nothrow_constructible_v<std::decay_t<T>, T>)
    {
        return std::optional<std::decay_t<T>>(std::forward<T>(t));
    }
```

Quuxplusone:
> How about making the first overload like this, which seems simpler to me?

Oh yeah — that works, because even though _today_ the first overload is used for e.g. `make_optional<int&>(i)`, there's no reason it needs to be. We can make the first overload non-viable in that case, and then the second overload will be used instead, which is fine! (Until someone points out a subtle ABI problem with that change, anyway... ;))

Updated accordingly.

steve-downey:
Closing, parts of this were included in the R8 proposal.

PR #64, opened 2024-09-25
Make `contiguous_iterator` constexpr iterator

### LINKED ISSUE #63: contiguous_iterator should be constexpr
```std::views::join``` on ```beman::optional26::optional``` doesn't work in constexpr context in my code https://godbolt.org/z/Ed11G66s3 (switch preprocessor if 1 to see error)

This code doesn't work in constexpr:
```cpp
auto range_result = view_of_optionals | std::views::join | std::views::take(100);
```

I expected it work like this C++23 code:
```cpp
auto range_result = view_of_optionals
        | std::views::filter([](const optional26<int>& opt) { return opt.has_value(); })
        | std::views::transform([](const optional26<int>& opt) { return *opt; })
        | std::views::take(100);
```
Filtering out all nullopt from view

With std::views::join compiler gives error that contiguous_iterator is not constexpr, so I simply tried editing it's declaration - marking everything I could as constexpr and afterwards this code worked as expected. So solely based on this I think it should be constexpr.

frederick-vs-ja:
Per [optional.iterators]/1 and the proposed [optionalref.iterators]/1 in P2988R7.

Rough test: https://godbolt.org/z/7eEhh954W

Fixes #63.

### DIFF
--- a/include/beman/optional26/detail/iterator.hpp
+++ b/include/beman/optional26/detail/iterator.hpp
@@ -43,18 +43,20 @@ struct contiguous_iterator : public base_contiguous_iterator<T, Container> {
     using typename base_type::reference;
 
     // Default constructor.
-    contiguous_iterator() noexcept : m_current() {}
+    constexpr contiguous_iterator() noexcept : m_current() {}
 
     // Pointer to iterator constructor.
-    contiguous_iterator(pointer it) noexcept : m_current(it) {}
+    constexpr contiguous_iterator(pointer it) noexcept : m_current(it) {}
 
     // As per P2727R4, for contiguous iterator we only need to provide operator*, operator+= and operator-.
-    reference operator*() const noexcept { return *m_current; }
-    auto&     operator+=(difference_type pos) noexcept {
+    constexpr reference operator*() const noexcept { return *m_current; }
+    constexpr auto&     operator+=(difference_type pos) noexcept {
         m_current += pos;
         return *this;
     }
-    difference_type operator-(contiguous_iterator other) const noexcept { return m_current - other.m_current; }
+    constexpr difference_type operator-(contiguous_iterator other) const noexcept {
+        return m_current - other.m_current;
+    }
 
   private:
     T* m_current;
--- a/src/beman/optional26/tests/detail/iterator.t.cpp
+++ b/src/beman/optional26/tests/detail/iterator.t.cpp
@@ -10,18 +10,27 @@
 #include <tests/test_types.hpp>
 
 #include <concepts>
+#include <cstdlib>
 #include <type_traits>
 #include <set>
 #include <vector>
 
-namespace beman::optional26::test {
+#define CONSTEXPR_EXPECT_EQ(val1, val2)   \
+    if (::std::is_constant_evaluated()) { \
+        if (!((val1) == (val2))) {        \
+            ::std::abort();               \
+        }                                 \
+    } else                                \
+        EXPECT_EQ(val1, val2)
+
+namespace beman::optional26::tests {
 
 // Dummy containers helper.
 struct container {};
 
-} // namespace beman::optional26::test
+} // namespace beman::optional26::tests
 
-using namespace beman::optional26::test;
+using namespace beman::optional26::tests;
 
 TEST(IteratorTest, IteratorConcepts) {
     const auto test = [](auto&& it) {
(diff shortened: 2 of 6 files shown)


include/beman/optional26/detail/iterator.hpp
```diff
@@ -43,18 +43,20 @@ struct contiguous_iterator : public base_contiguous_iterator<T, Container> {
     using typename base_type::reference;
 
     // Default constructor.
-    contiguous_iterator() noexcept : m_current() {}
+    constexpr contiguous_iterator() noexcept : m_current() {}
 
     // Pointer to iterator constructor.
-    contiguous_iterator(pointer it) noexcept : m_current(it) {}
+    constexpr contiguous_iterator(pointer it) noexcept : m_current(it) {}
```
frederick-vs-ja, line 49: This ctor currently allows implicit conversion, which in turn allows more permissive operations betweeen `contiguous_iterator` and raw pointer.
Should we disallow such implicit conversion, or even furtherly disallow explicit construction? (Perhaps should do in another PR.)

include/beman/optional26/detail/iterator.hpp
```diff
@@ -43,18 +43,20 @@ struct contiguous_iterator : public base_contiguous_iterator<T, Container> {
     using typename base_type::reference;
 
     // Default constructor.
-    contiguous_iterator() noexcept : m_current() {}
+    constexpr contiguous_iterator() noexcept : m_current() {}
```
neatudarius, line 46: I think these changes look fine.

@steve-downey @tzlaine , do you have any other comment?

neatudarius, line 46: I think these changes look fine.

@steve-downey @tzlaine @camio , do you have any other comment?

neatudarius, line 46: I also think we should revisit the test suite from https://github.com/beman-project/optional26/blob/main/src/beman/optional26/tests/detail/iterator.t.cpp and add some tests for the constexpr usage.

include/beman/optional26/detail/iterator.hpp
steve-downey: I've got a branch with updated wording and a number of fixes to conversion and dangling issues. It's exactly the current proposal. Getting the constraints and mandates to all play nice was a little tricky, but we did notice that assignment boiled down to one assignment from tuple, and that the move and copy assignments were identical, so we only need one. 

It is using the newish reference_converts_from_temporary the way that tuple does now. 

PR will be up later today.

steve-downey: The branch currently using aliases the traits that it uses into the Optional26 namespace in order to get the wording synopsis, constraints, mandates, effects all correct and in a form that I can just paste into the TeX doc for standardization, but probably shouldn't be done for the production version. 

It's also based on the pre renaming version. 

The really good news is that an LWG expert helped get it right, so this should be much smoother in the LEWG approval of wording and actual LWG review.

camio (commented):
Thanks for the contribution @frederick-vs-ja! This change looks good to me. Can we get a unit test that verifies the changes work as expected?

@steve-downey are you suggesting we hold off on merging this PR in anticipation of your bigger fix?

include/beman/optional26/detail/iterator.hpp
neatudarius: @frederick-vs-ja , also add one more entry into examples/

neatudarius:
@frederick-vs-ja , we would like to proceed with your PR. If new paper revision will change iterator interface at a later point, we'll deal with that in other PR.

can you applied the required feedback on this PR? 

CC: @camio @steve-downey , let's allow this improvement in main. Thanks @frederick-vs-ja for contribution into optional26!

include/beman/optional26/detail/iterator.hpp
frederick-vs-ja: The coverage for constexpr would be somehow large, I'll add it later today.

frederick-vs-ja: Added coverage in `optional_range_support.t.cpp` and `iterator.t.cpp`, but some stuffs don't work with GCC11. I'll check this soon.

src/beman/optional26/tests/optional_range_support.t.cpp
```diff
@@ -29,6 +29,10 @@
 #include <unordered_set>
 #include <vector>
 
+#if !defined(__clang__) && defined(__GNUC__) && (__GNUC__ < 12)
```
neatudarius, line 32: I don't like this solution. I would try to see exactly what gcc minor version we have on CI (you can add logs in yaml file).

Maybe it's worth considering upgrade to a newer gcc11 minor version or to remove gcc11 support. 

@steve-downey , do you have any input here?

frederick-vs-ja: I found that this is probably [GCC Bugzilla PR89074](https://gcc.gnu.org/PR89074). The bugfix landed in GCC 12.0 and wasn't backported to GCC 11 releases. So perhaps it's OK to only check the major version? (But perhaps the macro name should be more precise.)

steve-downey: I would be ok dropping 11 if we can't find a straightforward work around.

JeffGarland: My $.02 -- drop gcc 11 is circa  mid-2021. We're living on the cutting edge here.  I'd have a different opinion if all the later compilers didn't just handle this.

src/beman/optional26/tests/optional_range_support.t.cpp
```diff
@@ -91,96 +128,125 @@ TEST(RangeSupportTest, IteratorConcepts) {
...
+        test(beman::optional26::optional<derived>{derived{}});
     };
-
-    test(beman::optional26::optional<int>{26});
-    test(beman::optional26::optional<int*>{reinterpret_cast<int*>(0XCAFEBABE)});
-    test(beman::optional26::optional<empty>{empty{}});
-    test(beman::optional26::optional<no_default_ctor>{no_default_ctor{empty{}}});
-    test(beman::optional26::optional<base>{base{}});
-    test(beman::optional26::optional<derived>{derived{}});
+#if !defined(SKIP_CONSTANT_EVALUATION_BUG_FOR_GCC)
```
JeffGarland, line 182: Is this more gcc11 work around?

JeffGarland:
> @frederick-vs-ja , we would like to proceed with your PR. If new paper revision will change iterator interface at a later point, we'll deal with that in other PR.
> 
> can you applied the required feedback on this PR?
> 
> CC: @camio @steve-downey , let's allow this improvement in main. Thanks @frederick-vs-ja for contribution into optional26!

Concur here -- big steps forward on constexpr, which can be surprisingly tricky.

JeffGarland (approved):

src/beman/optional26/tests/optional_range_support.t.cpp
frederick-vs-ja: I'm now switching to remove GCC 11 coverage from CI, which should enable removing all such skippings.

src/beman/optional26/tests/detail/iterator.t.cpp
```diff
@@ -10,18 +10,27 @@
 #include <tests/test_types.hpp>
 
 #include <concepts>
+#include <cstdlib>
 #include <type_traits>
 #include <set>
 #include <vector>
 
-namespace beman::optional26::test {
+#define CONSTEXPR_EXPECT_EQ(val1, val2)   \
```
camio, line 18: I don't understand why this CONSTEXPR_ set of functions was introduced. Should we use `static_assert` be used to check the constexpr functionality?

src/beman/optional26/tests/optional_range_support.t.cpp
neatudarius: Cool. I'll make sure to remember this decision for exemplar.

src/beman/optional26/tests/test_types.hpp
```diff
@@ -50,7 +50,7 @@ struct derived : public base {
 
 struct move_detector {
     move_detector() = default;
-    move_detector(move_detector&& rhs) { rhs.been_moved = true; }
+    constexpr move_detector(move_detector&& rhs) { rhs.been_moved = true; }
```
neatudarius, line 53: Should we add constexpr to other/all constructors in this class/file?

src/beman/optional26/tests/detail/iterator.t.cpp
frederick-vs-ja: I'm already using `static_assert((lambda(), true));` now. These set of wrapping macros were introduced because gtest's corresponding macros are not constexpr-friendly.

neatudarius: First of all, this seems to be duplicated code from src/beman/optional26/tests/optional_range_support.t.cpp. So if we'll keep it, @frederick-vs-ja please move it in `est_utilities/test_types.*`.

2nd observation: I think we want to have all constexpr tests in a separate file, similar to https://github.com/beman-project/optional26/blob/main/src/beman/optional26/tests/optional_constexpr.t.cpp (optional.t.cpp vs optional_constexpr.t.cpp). @camio , @steve-downey ?

3rd: I think static_assert should work. IF not, this is how we do constexpr testing right now - https://github.com/beman-project/optional26/blob/main/src/beman/optional26/tests/optional_constexpr.t.cpp#L251-L263. Hope it hepls.

camio: > I'm already using static_assert((lambda(), true)); now. These set of wrapping macros were introduced because gtest's corresponding macros are not constexpr-friendly.

Oh, I see. You're doing this so you can use something like EXPECT_EQ from within a lambda evaluated at compile time. Cool.

src/beman/optional26/tests/test_types.hpp
steve-downey: The main reason for the duplication was that I duplicated all the tests to start with, and then worked to consteval them. It made it a little easier to track progress, but I wouldn't make it a mandatory pattern, necessarily. 
If it were consteval first, it might depend on implementation of the non-const version needs separate testing?

neatudarius (approved):
LGTM 👍

PR #69, opened 2024-10-05
Implement std::hash for beman::optional

steve-downey:
A std::hash implementation for optional, the hash of the underlying value, or 0 for disengaged.

### DIFF
--- a/src/beman/optional26/tests/optional.t.cpp
+++ b/src/beman/optional26/tests/optional.t.cpp
@@ -10,6 +10,7 @@
 #include <functional>
 #include <ranges>
 #include <tuple>
+#include <algorithm>
 
 #include <gtest/gtest.h>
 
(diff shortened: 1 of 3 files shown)


include/beman/optional26/optional.hpp
```diff
@@ -1147,4 +1147,21 @@ class optional<T&> {
...
 
+namespace std {
+template <typename T>
+    requires requires(T a) {
+        { std::hash<remove_const_t<T>>{}(a) } -> std::convertible_to<std::size_t>;
+    }
+struct hash<beman::optional26::optional<T>> {
+    static_assert(!is_reference_v<T>, "hash is not enabled for reference types");
+    size_t operator()(const beman::optional26::optional<T>& o) const
+        noexcept(noexcept(hash<remove_const_t<T>>{}(*o))) {
```
changkhothuychung, line 1158: Why do we need nested noexcept here. For example, is `noexcept(true)` different from `noexcept(noexcept(true))` ?

src/beman/optional26/tests/optional.t.cpp
```diff
@@ -596,6 +596,21 @@ TEST(OptionalTest, RangeTest) {
...
+    beman::optional26::optional<int> o3 = 42;
+    beman::optional26::optional<int> o4 = 42;
+
+    auto h1 = std::hash<beman::optional26::optional<int>>{}(o1);
+    auto h2 = std::hash<beman::optional26::optional<int>>{}(o2);
+    auto h3 = std::hash<beman::optional26::optional<int>>{}(o3);
+    auto h4 = std::hash<beman::optional26::optional<int>>{}(o4);
+
+    EXPECT_EQ(h1, h2);
+    EXPECT_EQ(h3, h4);
```
neatudarius, line 611: I would also add `EXPECT_NEQ(h1, h3)` in this test and creata new one where I would expect that hashing of a series of number inside optional gives different hash values - e.g. 1, 2, 4, 8, 16, ...

src/beman/optional26/tests/optional.t.cpp
```diff
@@ -596,6 +596,21 @@ TEST(OptionalTest, RangeTest) {
     }
 }
 
+TEST(OptionalTest, HashTest) {
```
neatudarius, line 599: Maybe we should move this test at the end of this file? I'm not expecting to see hashing tests before tests for contructors or other basic features.

steve-downey: It's before the tests that I imported from views maybe, but since that doesn't really make sense for anyone but me, I'll rename those tests and slot the hash in later.

neatudarius (approved):

PR #70, opened 2024-10-12
R7 and 8 of the paper and fixes to optional after wording and LEWG reviews. 

steve-downey:
Wording review for paper, and fixes to make sure the constraints/mandates enforce the non-dangling design and other improvements in the API as suggested in reviews.


neatudarius (commented):
LGTM 👍

PR #71, opened 2024-10-22
Adopt lint configuration from `exemplar` project

steve-downey:
Use better supported tools.
Fix lint issues noted - except for some src block and references.

### DIFF
--- a/.markdownlint.json
+++ b/.markdownlint.json
@@ -1,4 +1,0 @@
-{
-    "MD013": false,
-    "MD024": false
-}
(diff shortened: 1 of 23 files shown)


camio:
It looks like the markdown linter is finding some issues in the `README.md`

steve-downey:
Yes, unfortunately two in long commands in a src block of shell script, and one in the references so fixing it isn't as trivial as fill-paragraph. 

Most of the others were easy to fix, or suppress.

camio (approved):

PR #73, opened 2024-11-09
Install sets

steve-downey:
This changes the install and include  directory strategy to use FILE_SETs with type HEADERS. 
Moves the creation of the targets and named header file sets upward so they exist in order to add the files locally in the src and include directories. Change the installs to use components so we can install just the necessary headers and not the empty library. 

Also make the option to build headers named for the project so it doesn't affect other included projects.

Exclude google from all so it doesn't get installed for the "all" component. 

Remove the cmake export install, at least for now. It was not working correctly. It's probably not strictly necessary for vcpkg or conan integration.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -12,11 +12,16 @@ include(FetchContent)
 
 set(TARGETS_EXPORT_NAME ${CMAKE_PROJECT_NAME}Targets)
 
-# Build the tests only if enabled via the CLI flag: BUILD_TESTING.
-if(BUILD_TESTING)
+option(OPTIONAL26_ENABLE_TESTING
+  "Enable building tests and test infrastructure"
+  ${PROJECT_IS_TOP_LEVEL})
+
+# Build the tests if enabled via the option OPTIONAL26_ENABLE_TESTING
+if(OPTIONAL26_ENABLE_TESTING)
     # Fetch GoogleTest
     FetchContent_Declare(
         googletest
+        EXCLUDE_FROM_ALL
         GIT_REPOSITORY https://github.com/google/googletest.git
         GIT_TAG
             e39786088138f2749d64e9e90e0f9902daa77c40 # release-1.15.0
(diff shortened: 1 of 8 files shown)


CMakeLists.txt
```diff
@@ -12,51 +12,49 @@ include(FetchContent)
...
 )
 
-include(CMakePackageConfigHelpers)
-
-write_basic_package_version_file(
-    ${CMAKE_CURRENT_BINARY_DIR}/${CMAKE_PROJECT_NAME}ConfigVersion.cmake
-    VERSION ${PROJECT_VERSION}
-    COMPATIBILITY AnyNewerVersion
-)
+if(OPTIONAL26_ENABLE_TESTING)
```
neatudarius, line 41: Can you remind me why BUILD_TESTING is not enough/OK ?

CMakeLists.txt
```diff
@@ -12,51 +12,49 @@ include(FetchContent)
...
 
-# Build the tests only if enabled via the CLI flag: BUILD_TESTING.
-if(BUILD_TESTING)
+option(
+    OPTIONAL26_ENABLE_TESTING
+    "Enable building tests and test infrastructure"
+    ${PROJECT_IS_TOP_LEVEL}
+)
+
+# Build the tests if enabled via the option OPTIONAL26_ENABLE_TESTING
```
neatudarius, line 21: Can you remind me why BUILD_TESTING is not enough/OK ?

The Beman Standard specify to have `BUILD_TESTING`.

steve-downey: It doesn't compose. Best practice for Cmake is to prefix options with the project they apply to, and to default the value to true for a top level project. 

Patten documented in Professional CMake.

neatudarius: Agree that it's the best practice from CMake docs, but I would expect to first make a PR in the Beman Standard.

neatudarius (approved):

CMakeLists.txt
steve-downey: Fortunately, it's a recommendation, but I'll PR that, now that I've seen it. But since this actually breaks trying to compose this into another beman project, it would be nice to fix it now.

neatudarius: @JeffGarland should we make it a requirement?

steve-downey:
It's probably a better requirement than producing the Cmake file for
FindPackage or ExternalProject use. I don't think we have any evidence that
we're doing that correctly?

PR #75, opened 2024-11-10
Verify headers

steve-downey:
Convert to an INTERFACE target rather than static library and use the new-ish cmake verify headers mechanism to test that the headers compile and are stand alone.


neatudarius (approved):
Interesting. I think we should add this idea to our tomorrow's agenda.

I'm what others think. TLDR: We should have a rule in the standard for header only libraries!

steve-downey:
It's a relatively new Cmake feature, but getting rid of the static library
is a win.
College tour tomorrow, so I'll miss the meeting, but please do bring it up.

include/beman/optional26/CMakeLists.txt
```diff
@@ -13,3 +13,29 @@ target_sources(
...
+    ${CMAKE_INSTALL_LIBDIR}
+    FILE_SET beman_optional26_headers
+        DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}
+        COMPONENT beman_optional26_development
+)
+
+install(
+    EXPORT beman_optional26_export
+    DESTINATION ${CMAKE_INSTALL_LIBDIR}/cmake/beman/optional26/
+    NAMESPACE Beman::Optional26::
```
neatudarius, line 37:
```suggestion
    NAMESPACE beman::optional26::
```
?

CMakeLists.txt
```diff
@@ -31,11 +31,13 @@ if(OPTIONAL26_ENABLE_TESTING)
     FetchContent_MakeAvailable(googletest)
 endif()
 
+set(CMAKE_VERIFY_INTERFACE_HEADER_SETS ON)
+
 # Create the library target and named header set for beman_optional26
-add_library(beman_optional26 STATIC)
+add_library(beman_optional26 INTERFACE)
```
neatudarius, line 37:
```suggestion
add_library(beman.optional26 INTERFACE)
```
The standard says it should use `.` after beman. Would you like to apply these updates in your PR?

image

neatudarius:
@steve-downey , please proceed with this PR. We discussed yesterday in the sync and it's OK to propose this standard cmake practice in the Beman standard (same document, a new CMAKE.$NAME rule). Can you add this recommendation?

CMakeLists.txt
steve-downey: I can fix this up as part of the PR. Need to merge in other changes in any case.

neatudarius: That would be great. But feel free to speed-up/skip steps today if you need something in main for Poland.

steve-downey:
> @steve-downey , please proceed with this PR. We discussed yesterday in the sync and it's OK to propose this standard cmake practice in the Beman standard (same document, a new CMAKE.$NAME rule). Can you add this recommendation?
> 
> If you cannot help with that, I can do it after Poland week (in 2-3 weeks).

Names of the header set, install component, or all of the above? 
The header set name needs to be non-colliding to allow add subdir composition. If it's not clear in the standard that the goal is to allow project composition, we should do so, just to clarify that if you find something that conflicts with composition, fixing it is allowed over rote adherence to the standard (not an issue here...) 
The component name is part of the interface to the project, so it's probably even more important to standardise that. In particular it's likely to affect how building within vcpkg and conan, etc work.

neatudarius:
> > @steve-downey , please proceed with this PR. We discussed yesterday in the sync and it's OK to propose this standard cmake practice in the Beman standard (same document, a new CMAKE.$NAME rule). Can you add this recommendation?
> > If you cannot help with that, I can do it after Poland week (in 2-3 weeks).
> 
> Names of the header set, install component, or all of the above? The header set name needs to be non-colliding to allow add subdir composition. If it's not clear in the standard that the goal is to allow project composition, we should do so, just to clarify that if you find something that conflicts with composition, fixing it is allowed over rote adherence to the standard (not an issue here...) The component name is part of the interface to the project, so it's probably even more important to standardise that. In particular it's likely to affect how building within vcpkg and conan, etc work.

It sounds like it's a good idea to put in standard all of the above. I would approve that, and David was also happy about your PR. It's true that we had only 5 people in the last sync. If you want to get more feedback, do a discourse thread, but I would go directly with the PR in BEMAN_STANDARD.md.

CMakeLists.txt
ClausKlein: Note: that will end in very strange behaviour! Normally I use something like this:
```cmake
find_package(Boost CONFIG QUIET)
if(TARGET Boost::headers AND TARGET fmt::fmt)
    add_executable(to_string to_string.cpp)
    target_link_libraries(to_string fmt::fmt Boost::headers)
endif()
```

ClausKlein:
this ends in this `cmake export packages` which is not usable!
```
bash-5.2$ tree .install/clang-19/lib/cmake/
.install/clang-19/lib/cmake/
`-- beman
    `-- optional26
        |-- BemanOptional26Config.cmake
        `-- beman_optional26.cmake

3 directories, 2 files
bash-5.2$ 
```

Also the `packagename-version.cmake` is missing.

ClausKlein:
A `header only library `is architecture and compiler independent.

Why install under `prefix/clang-19`?

ClausKlein:
```bash
bash-5.2$ cmake -S . -G Ninja -B .build -D CMAKE_PREFIX_PATH=../.install/clang-19
CMake Error at CMakeLists.txt:13 (find_package):
  Could not find a configuration file for package "beman_optional26" that
  exactly matches requested version "0.0.0".

  The following configuration files were considered but not accepted:

    /Users/clausklein/Workspace/cpp/beman-project/optional26/.install/clang-19/lib/cmake/beman_optional26/beman_optional26-config.cmake, version: unknown

-- Configuring incomplete, errors occurred!
bash-5.2$ 
```

CMakeLists.txt
steve-downey: This is creating the target that gets the named header file_set added to it? It's up here to avoid ordering problems between subdirectories.

steve-downey:
> A `header only library `is architecture and compiler independent.
> 
> Why install under `prefix/clang-19`?

Habit, I suppose? Easier to always disambiguate than sort out of its strictly necessary. 
Although if there are any other components of the project installed, not having collisions is useful. 
In my normal usage, this  install location is transient, and just used to produce a deb that's installed later.

steve-downey:
I'll work on the installation bugs and catch up with the Beman Standard in follow on PRs.

PR #80, opened 2024-11-16
Project requires at least C++20

autoantwort:


camio:
We haven't been using `target_compile_features` because doing so can lead to subtle ODR violations in large projects. See https://github.com/beman-project/exemplar/issues/40#issuecomment-2388995030 for a more elaborated rationale.

autoantwort:
Hm, ideally the "large project" would also use cmake to links it stuff together and then the c++20 flag would get propagated to the final executable. 
In vcpkg CMAKE_CXX_STANDARD don't gets set by default (the users can if they want), so it currently simply fails to build since the default one is c++11 for apple clang.

camio:
> Hm, ideally the "large project" would also use cmake to links it stuff together and then the c++20 flag would get propagated to the final executable.

Large projects aren't built completely from source. Frequently, they incorporate shared libraries that come as binaries which exacerbates the ODR violation issues.

My understanding is that VCPkg interacts with CMake by providing a toolchain file. How does this toolchain file determine compiler flags?

steve-downey:
Changing the flags of a consumer is a great way to break the consumer, with the breakage varying between subtle and loud. Loud is good because it gets reverted quickly. 
Even if I'm building all of my project from source, having a dependency make a huge change, like std level, can be a problem. 
Saying "No" and producing an error might be better, but the compiler errors should be fairly clear even without that.

ClausKlein:
I do not see that this may a problem for at least a `header only library`:

https://gitlab.kitware.com/cmake/cmake/-/issues/19136#note_1598283

ClausKlein:
This seems not a better way:
```bash
bash-5.2$ git ls-files ::*.cmake ::*CMakeLists.txt ::*.json | xargs egrep --color -rw '(CMAKE_CXX_EXTENSIONS|CMAKE_CXX_STANDARD\w*|std)' 
CMakePresets.json:        "CMAKE_CXX_EXTENSIONS": false,
CMakePresets.json:        "CMAKE_CXX_STANDARD": "20",
CMakePresets.json:        "CMAKE_CXX_STANDARD_REQUIRED": true,
etc/ci-clang-toolchain.cmake:    "-std=c++20 \
etc/clang-flags.cmake:set(CMAKE_CXX_STANDARD 20)
etc/gcc-flags.cmake:set(CMAKE_CXX_STANDARD 23)
etc/gcc-flags.cmake:set(CMAKE_CXX_FLAGS "-std=c++23 -Wall -Wextra " CACHE STRING "CXX_FLAGS" FORCE)
etc/gcc-toolchain.cmake:    "-std=c++20 \
etc/llvm-16-toolchain.cmake:    "-std=c++20 \
etc/llvm-master-toolchain.cmake:    "-std=c++2a \
etc/llvm-toolchain.cmake:    "-std=c++20 \
src/beman/optional26/tests/CMakeLists.txt:            "-DCMAKE_CXX_STANDARD=${CMAKE_CXX_STANDARD}"
```

neatudarius (changes requested):
As per ps://github.com/bemanproject/beman/pull/76, 

"Therefore, `target_compile_features` (e.g., `cxx_std_20`) must not be used
because it modifies the compilation environment of dependent targets. Compiler
support for required features should be determined at CMake configuration time
using `check_cxx_source_compiles`."


Please hold this PR until we finish updating the Beman Standard with this missing piece.

ClausKlein:
### The project is completely on a wrong way,
 please read https://crascit.com/professional-cmake/ before make suche a decision!

camio:
> ### The `beman_project` is partly on a realy wrong way,

@ClausKlein, please review our [code of conduct](https://github.com/bemanproject/beman/blob/main/docs/CODE_OF_CONDUCT.md). This comment does not embody the  friendliness, patience, and respect that we adhere to in these spaces.

> please read https://crascit.com/professional-cmake/ before make suche a decision!

Telling someone to read a book is not a reasonable way to move a conversation forward.

ClausKlein:
OK ["Das Gras wächst nicht schneller, wenn man daran zieht"](https://www.sg.ch/news/sgch_bildungsdepartement/2021/10/-das-gras-waechst-nicht-schneller--wenn-man-daran-zieht--.html#:~:text=Dieses%20afrikanische%20Sprichwort%2C%20welches%20Remo,nicht%20die%20Entwicklung%20direkt%20beeinflussen.)

neatudarius:
@ClausKlein @camio , is this PR still applicable after https://github.com/bemanproject/beman/pull/76?

camio:
> @ClausKlein @camio , is this PR still applicable after https://github.com/bemanproject/beman/pull/76?

No, this approach violates `[CMAKE.PASSIVE_TARGETS]`.

neatudarius:
@autoantwort , would you like to propose other solution for this issue? The solution is not valid as it is not conformant with [CMAKE.PASSIVE_TARGETS], but indeed the issue is valid , as state in root README: [CMAKE.PASSIVE_TARGETS]

PR #85, opened 2024-12-07
Fix builds on macOS

ClausKlein:
Also applied a `pre-commit autoupdate `

this fix
- https://github.com/bemanproject/optional26/issues/86
- https://github.com/bemanproject/optional26/issues/84

### DIFF
--- a/etc/gcc-flags.cmake
+++ b/etc/gcc-flags.cmake
@@ -28,12 +28,21 @@ set(CMAKE_CXX_FLAGS_TSAN
     "C++ TSAN Flags"
     FORCE
 )
-set(CMAKE_CXX_FLAGS_ASAN
-    "-O3 -g -DNDEBUG -fsanitize=address,undefined,leak"
-    CACHE STRING
-    "C++ ASAN Flags"
-    FORCE
-)
+if(APPLE)
+    set(CMAKE_CXX_FLAGS_ASAN
+        "-O3 -g -DNDEBUG -fsanitize=address,undefined"
+        CACHE STRING
+        "C++ ASAN Flags"
+        FORCE
+    )
+else()
+    set(CMAKE_CXX_FLAGS_ASAN
+        "-O3 -g -DNDEBUG -fsanitize=address,undefined,leak"
+        CACHE STRING
+        "C++ ASAN Flags"
+        FORCE
+    )
+endif()
 
 set(CMAKE_CXX_FLAGS_GCOV
     "-O0 -fno-inline -g --coverage -fprofile-abs-path"
--- a/include/beman/optional26/detail/stl_interfaces/iterator_interface.hpp
+++ b/include/beman/optional26/detail/stl_interfaces/iterator_interface.hpp
@@ -229,7 +229,7 @@ BEMAN_OPTIONAL26_DETAIL_STL_INTERFACES_NAMESPACE_V1 {
         template <typename D = Derived>
         constexpr auto operator->() const
             noexcept(noexcept(detail::make_pointer<pointer, reference>(*std::declval<const D&>())))
-            -> decltype(detail::make_pointer<pointer, reference>(*std::declval<const D&>())) {
+                -> decltype(detail::make_pointer<pointer, reference>(*std::declval<const D&>())) {
             return detail::make_pointer<pointer, reference>(*derived());
         }
 
--- a/include/beman/optional26/detail/stl_interfaces/iterator_interface.hpp
+++ b/include/beman/optional26/detail/stl_interfaces/iterator_interface.hpp
@@ -229,7 +229,7 @@ BEMAN_OPTIONAL26_DETAIL_STL_INTERFACES_NAMESPACE_V1 {
         template <typename D = Derived>
         constexpr auto operator->() const
             noexcept(noexcept(detail::make_pointer<pointer, reference>(*std::declval<const D&>())))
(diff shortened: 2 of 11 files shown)


etc/gcc-flags.cmake
steve-downey, line 1: I'd prefer to keep the toolchain file declarative and split out appleclang into a separate toolchain.

include/beman/optional26/detail/stl_interfaces/iterator_interface.hpp
```diff
@@ -229,7 +229,7 @@ BEMAN_OPTIONAL26_DETAIL_STL_INTERFACES_NAMESPACE_V1 {
         template <typename D = Derived>
         constexpr auto operator->() const
             noexcept(noexcept(detail::make_pointer<pointer, reference>(*std::declval<const D&>())))
-            -> decltype(detail::make_pointer<pointer, reference>(*std::declval<const D&>())) {
+                -> decltype(detail::make_pointer<pointer, reference>(*std::declval<const D&>())) {
```
steve-downey, line 232: If this is from a reformat, it should be applied as a distinct change and the ignore list updated.

papers/P2988/optional.hpp
```diff
@@ -326,8 +326,8 @@ constexpr auto optional<T&>::and_then(F&& f) const {
 
 template <class T>
 template <class F>
-constexpr auto
-optional<T&>::transform(F&& f) const -> optional<invoke_result_t<F, T&>> {
+constexpr auto optional<T&>::transform(F&& f) const
+    -> optional<invoke_result_t<F, T&>> {
```
steve-downey, line 330: This is for the paper, and should be identical, using the draft standard format, if the new clang-format is changing it, it may need tweaking.

etc/gcc-flags.cmake
ClausKlein: `Clang-19` on `Windows`,  `Linux`, and `Darwin` my be used with the same preset.
The only differences are in setting sanitiser flags.

Wy duplicate 95 % of cmake code for each OS?

include/beman/optional26/detail/stl_interfaces/iterator_interface.hpp
ClausKlein: reverted

papers/P2988/optional.hpp
ClausKlein: reverted

etc/gcc-flags.cmake
ClausKlein: OK, we may use an other implementation:
```cmake
option(ENABLE_ASAN "Enable AddressSanitizer" YES)

if(MSVC)
  if(ENABLE_ASAN)
    string(REPLACE "/RTC1" "" CMAKE_C_FLAGS_DEBUG "${CMAKE_C_FLAGS_DEBUG}")
    string(REPLACE "/RTC1" "" CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG}")
    add_compile_options(/fsanitize=address /fsanitize-address-use-after-return)
  endif()
elseif(CMAKE_C_COMPILER_ID MATCHES "GNU|Clang")

  option(ENABLE_LSAN "Enable LeakSanitizer" NO)
  option(ENABLE_TSAN "Enable ThreadSanitizer" NO)
  option(ENABLE_UBSAN "Enable UndefinedBehaviorSanitizer" YES)
  if(NOT APPLE)
    option(ENABLE_MSAN "Enable MemorySanitizer" NO)
  endif()
  if((ENABLE_ASAN AND (ENABLE_TSAN OR ENABLE_MSAN)) OR
     (ENABLE_LSAN AND (ENABLE_TSAN OR ENABLE_MSAN)) OR
     (ENABLE_TSAN AND ENABLE_MSAN))
     message(FATAL_ERROR
       "Invalid sanitizer combination:\n"
       "  ENABLE_ASAN:  ${ENABLE_ASAN}\n"
       "  ENABLE_LSAN:  ${ENABLE_LSAN}\n"
       "  ENABLE_TSAN:  ${ENABLE_TSAN}\n"
       "  ENABLE_MSAN:  ${ENABLE_MSAN}"
     )
  endif()
  add_compile_options(
         -fno-omit-frame-pointer
         $<$<BOOL:${ENABLE_ASAN}>:-fsanitize=address>
         $<$<BOOL:${ENABLE_LSAN}>:-fsanitize=leak>
         $<$<BOOL:${ENABLE_MSAN}>:-fsanitize=memory>
         $<$<BOOL:${ENABLE_TSAN}>:-fsanitize=thread>
         $<$<BOOL:${ENABLE_UBSAN}>:-fsanitize=undefined>
  )
  add_link_options(
         $<$<BOOL:${ENABLE_ASAN}>:-fsanitize=address>
         $<$<BOOL:${ENABLE_LSAN}>:-fsanitize=leak>
         $<$<BOOL:${ENABLE_MSAN}>:-fsanitize=memory>
         $<$<BOOL:${ENABLE_TSAN}>:-fsanitize=thread>
         $<$<BOOL:${ENABLE_UBSAN}>:-fsanitize=undefined>
  )
endif()
```

neatudarius: I would also want separate files, as @steve-downey suggestef

neatudarius (changes requested):
Please address feedback. Thanks for contributions!

ClausKlein:
No, I will not continue this endless waiting for constructive waiting.

PR #88, opened 2024-12-10
Update pre-commit configs

wusatosi:
`.shellcheckrc`, `.codespell_ignore` and `.cmake-format.json` is currently unused due to #71 .

`.cmake-format.json` is not used by `gersemi`, the current cmake linter, thus removed.
`.shellcheckrc` is not used as shellcheck was removed in #71 . This is removed as there isn't significant shell scripts in this repo.
codespell is added back, this suggestion is up to the current maintainers to remove/ add back. I will add codespell to upstream (exemplar) as well, I don't know why I didn't add this in the first place.

Dependencies of pre-commit hooks are updated as well.


steve-downey (approved):
LGTM

PR #90, opened 2024-12-11
Fix git blame ignore for GitHub

wusatosi:
Current GitHub blame doesn't work.
image

Seem like a full commit hash is needed to play well with GitHub.


steve-downey (approved):
+1

PR #93, opened 2024-12-31
When moving a 'optional<T&>', don't move the remote value

### LINKED ISSUE #92: What should moving from `optional<T&>` into `optional<T>` do?
I'm wondering about the interactions between `optional<T&>` and `optional<T>`. Specifically about `optional<T>`'s "converting" move constructor when the RHS is a `optional<T&>`. Should there be a move from the "remote" value, or should the value category be shallow, leading to a copy? Example:

```c++
    {
        std::string s = "qwertyuiopasdfghjklzxcvbnmqwertyuiopasdfghjklzxcvbnm";
        beman::optional26::optional<std::string&> o1 = s;
        beman::optional26::optional<std::string> o2 = std::move(o1);
        EXPECT_???(s.empty());
    }
```

At the moment, there is a move happening, and `s` is empty. I'm not sure if this is intended behavior though, or a "happy accident".

In any case, the "converting" move constructor of `optional<T>` seems not yet aware of `optional<T&>`:

```c++
/// Converting move constructor.
template <class T>
template <class U>
inline constexpr optional<T>::optional(optional<U>&& rhs)
    requires detail::enable_from_other<T, U, U&&> && std::is_convertible_v<U&&, T>
{
    if (rhs.has_value()) {
        construct(std::move(*rhs));
    }
}
```

...if `U` is a reference here, the `U&&`'s collapse to simple references. So `detail::enable_from_other` actually checks if it can _copy_ the value of the `rhs` in the constraints, but then does a `move` inside the function body.

jiixyj:
Moving `optional<T&>` into `optional<T>` should probably not result in a move of the remote value. Because of reference collapsing, the constraints of the move constructor/assignment operator of `optional<T>` already check for a copy of the remote value, not a move.

This makes the code match the constraints by ~~moving the `move` before the dereference, since for `optional<T&>`, `*std::move(rhs)` always results in a lvalue reference, not a rvalue reference.~~ introducing new constructor/assignment overloads for `optional<T&>`.

edit: fixes #92
edit: updated for new approach

### DIFF
--- a/include/beman/optional26/optional.hpp
+++ b/include/beman/optional26/optional.hpp
@@ -488,7 +488,7 @@ inline constexpr optional<T>::optional(optional<U>&& rhs)
     requires detail::enable_from_other<T, U, U&&> && std::is_convertible_v<U&&, T>
 {
     if (rhs.has_value()) {
-        construct(std::move(*rhs));
+        construct(*std::move(rhs));
     }
 }
 
@@ -498,7 +498,7 @@ inline constexpr optional<T>::optional(optional<U>&& rhs)
     requires detail::enable_from_other<T, U, U&&> && (!std::is_convertible_v<U &&, T>)
 {
     if (rhs.has_value()) {
-        construct(std::move(*rhs));
+        construct(*std::move(rhs));
     }
 }
 
@@ -594,14 +594,14 @@ inline constexpr optional<T>& optional<T>::operator=(optional<U>&& rhs)
 {
     if (has_value()) {
         if (rhs.has_value()) {
-            value_ = std::move(*rhs);
+            value_ = *std::move(rhs);
         } else {
             hard_reset();
         }
     }
 
     else if (rhs.has_value()) {
-        construct(std::move(*rhs));
+        construct(*std::move(rhs));
     }
 
     return *this;
--- a/include/beman/optional26/optional.hpp
+++ b/include/beman/optional26/optional.hpp
@@ -294,11 +294,11 @@ class optional {
 
     template <class U>
     constexpr explicit(!std::is_convertible_v<U&, T>) optional(const optional<U&>& rhs)
-        requires(detail::enable_from_other<T, U, U&> && std::is_convertible_v<U&, T>);
+        requires(detail::enable_from_other<T, U&, U&> && std::is_convertible_v<U&, T>);
 
     template <class U>
     constexpr explicit(!std::is_convertible_v<U&, T>) optional(const optional<U&>& rhs)
-        requires(detail::enable_from_other<T, U, U&> && !std::is_convertible_v<U&, T>);
+        requires(detail::enable_from_other<T, U&, U&> && !std::is_convertible_v<U&, T>);
 
     // \ref{optional.dtor}, destructor
     constexpr ~optional()
@@ -341,7 +341,7 @@ class optional {
 
     template <class U>
     constexpr optional& operator=(const optional<U&>& rhs)
-        requires(detail::enable_assign_from_other<T, U, U&>);
+        requires(detail::enable_assign_from_other<T, U&, U&>);
 
     template <class... Args>
     constexpr T& emplace(Args&&... args);
@@ -521,7 +521,7 @@ inline constexpr optional<T>::optional(optional<U>&& rhs)
 template <class T>
 template <class U>
 inline constexpr optional<T>::optional(const optional<U&>& rhs)
-    requires(detail::enable_from_other<T, U, U&> && std::is_convertible_v<U&, T>)
+    requires(detail::enable_from_other<T, U&, U&> && std::is_convertible_v<U&, T>)
 {
     if (rhs.has_value()) {
         construct(*rhs);
@@ -531,7 +531,7 @@ inline constexpr optional<T>::optional(const optional<U&>& rhs)
 template <class T>
 template <class U>
 inline constexpr optional<T>::optional(const optional<U&>& rhs)
-    requires(detail::enable_from_other<T, U, U&> && !std::is_convertible_v<U&, T>)
+    requires(detail::enable_from_other<T, U&, U&> && !std::is_convertible_v<U&, T>)
 {
     if (rhs.has_value()) {
         construct(*rhs);
@@ -646,7 +646,7 @@ inline constexpr optional<T>& optional<T>::operator=(optional<U>&& rhs)
 template <class T>
 template <class U>
 inline constexpr optional<T>& optional<T>::operator=(const optional<U&>& rhs)
-    requires(detail::enable_assign_from_other<T, U, U&>)
+    requires(detail::enable_assign_from_other<T, U&, U&>)
 {
     if (has_value()) {
         if (rhs.has_value()) {
--- a/include/beman/optional26/optional.hpp
+++ b/include/beman/optional26/optional.hpp
@@ -276,19 +276,29 @@ class optional {
 
     template <class U>
     constexpr explicit(!std::is_convertible_v<U, T>) optional(const optional<U>& rhs)
-        requires detail::enable_from_other<T, U, const U&> && std::is_convertible_v<const U&, T>;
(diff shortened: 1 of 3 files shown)


src/beman/optional26/tests/optional.t.cpp
```diff
@@ -868,3 +868,42 @@ TEST(OptionalTest, HashTest) {
         EXPECT_EQ(h1, h2);
     }
 }
+
+// Moving an `optional<T&>` should not move the remote value.
+// Check this by deleting the rvalue overloads of some (unrealistic) types.
+TEST(OptionalTest, MoveRef) {
+    struct copyable_but_not_movable {
```
neatudarius, line 875: same obersation - we have a header file with test classes in src/beman/optional26/tests/test_types.hpp

include/beman/optional26/optional.hpp
```diff
@@ -488,7 +488,7 @@ inline constexpr optional<T>::optional(optional<U>&& rhs)
     requires detail::enable_from_other<T, U, U&&> && std::is_convertible_v<U&&, T>
 {
     if (rhs.has_value()) {
```
steve-downey, line 490: https://eel.is/c++draft/optional#ctor-34 means this needs a paper.

src/beman/optional26/tests/optional.t.cpp
```diff
@@ -868,3 +868,42 @@ TEST(OptionalTest, HashTest) {
         EXPECT_EQ(h1, h2);
     }
 }
+
```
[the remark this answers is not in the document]
jiixyj: > With beman::optional, this also fails to compile. The title of the PR lead me to think this was doing the wrong thing successfully. Do you have a case where moving an optional<T&> incorrectly moves the referred to value?

Yes, I have a more realistic example at the issue: <https://github.com/bemanproject/optional26/issues/92>

include/beman/optional26/optional.hpp
jiixyj: Yep, and there is https://eel.is/c++draft/optional#assign-25 as well.

Although in a world where only `optional<T>` exists I believe `std::move(*rhs)` and `*std::move(rhs)` do the same thing. So implementers could have used the as-if rule this whole time. Only for `optional<T&>` there might be a semantic difference. Maybe P2988 is a good vehicle for this?

src/beman/optional26/tests/optional.t.cpp
jiixyj: Thanks for the pointer! I'll update the PRs accordingly.

steve-downey:
If it's that a optional<Foo> will steal the referred to of optional<Foo&>
in the converting assignment, it might be safer to specialize the
assignment operator? If U is a reference it should always copy assign
regardless of the value category of the optional.

jiixyj:
> If it's that a optional<Foo> will steal the referred to of optional<Foo&> in the converting assignment, it might be safer to specialize the assignment operator? If U is a reference it should always copy assign regardless of the value category of the optional.

This sounds sensible to me! It occured to me that the constraints are slightly wrong even for the `optional(const optional<U>& rhs)` converting constructor when called with a `optional<U&>`, so having some overloads for `optional<T&>` would be more correct. Did you have something like the following in mind?

```c++

    template <class U>
    constexpr explicit(!std::is_convertible_v<U, T>) optional(const optional<U>& rhs)
        requires(!std::is_reference_v<U> && detail::enable_from_other<T, U, const U&> && std::is_convertible_v<const U&, T>);
    template <class U>
    constexpr explicit(!std::is_convertible_v<U, T>) optional(const optional<U>& rhs)
        requires(!std::is_reference_v<U> && detail::enable_from_other<T, U, const U&> && !std::is_convertible_v<const U&, T>);

    template <class U>
    constexpr explicit(!std::is_convertible_v<U, T>) optional(optional<U>&& rhs)
        requires(!std::is_reference_v<U> && detail::enable_from_other<T, U, U &&> && std::is_convertible_v<U &&, T>);
    template <class U>
    constexpr explicit(!std::is_convertible_v<U, T>) optional(optional<U>&& rhs)
        requires(!std::is_reference_v<U> && detail::enable_from_other<T, U, U &&> && !std::is_convertible_v<U &&, T>);

    template <class U>
    constexpr explicit(!std::is_convertible_v<U, T>) optional(const optional<U&>& rhs)
        requires(detail::enable_from_other<T, U, U&> && std::is_convertible_v<U&, T>);
    template <class U>
    constexpr explicit(!std::is_convertible_v<U, T>) optional(const optional<U&>& rhs)
        requires(detail::enable_from_other<T, U, U&> && !std::is_convertible_v<U&, T>);
```

...this adds `!std::is_reference_v<U>` constraints to the `optional<U>` overloads, and the constraints for `optional<U&>` check for constructibility from `U&` (without `const`, since `const` is shallow).

steve-downey:
Or using an `if constexpr` to change the body of the affected assignment
operators, as is_reference doesn't affect participation in overload, just
what the 'Effects' are for the assignment, where for an optional<U&> it
should always treat the U as an lvalue, which I think is what
*std::move(rhs) ended up doing, but by more complicated machinery. The cv
qualifiers and value category of the optional are shallow, so not relevant,
so there might be only one overload necessary on the reference side?

I think having the explicit !std::is_reference_v<U> and
std::is_reference_v<U> also avoids subsumption problems in the compiler in
checking the requires clauses, which is partly why my inclination would be
to if constexpr in the body.

jiixyj:
I've implemented the approach with the separated overloads for `optional<U&>`. And added some more exhaustive tests to really make sure the desired operations on the `U&` are called. It is now nicely symmetrical compared to the monadic operations that also always operate on the `U&` regardless of the value category of the `optional<U&>`.

> Or using an `if constexpr` to change the body of the affected assignment operators

Can you elaborate? I'm not sure how this would look like, as the constraints are different for all the overloads, especially the `detail::enable_from_other`. For example it would be bad if the `detail::enable_from_other` checked constructibility from `const U&`, but then inside the body one would construct from `U&`.

steve-downey:
That sounds like it might not work, then, although I think that's also a
problem for the initial approach because the result of the optional deref
operation can be different? What you've described sounds like a much better
approach, though.

It does mean I'll be busy in the next week to update P2988.

steve-downey:
A simple set of failing test cases
```
TEST(OptionalTest, TestStealingFromReference) {
    beman::optional26::tests::copyable_move_detector                               m;
    beman::optional26::optional<beman::optional26::tests::copyable_move_detector&> om{m};
    beman::optional26::optional<beman::optional26::tests::copyable_move_detector>  o{std::move(om)};
    EXPECT_FALSE(m.been_moved);
}

TEST(OptionalTest, TestAssignStealingFromReference) {
    beman::optional26::tests::copyable_move_detector                               m;
    beman::optional26::optional<beman::optional26::tests::copyable_move_detector&> om{m};
    beman::optional26::optional<beman::optional26::tests::copyable_move_detector>  o;
    o = std::move(om);
    EXPECT_FALSE(m.been_moved);
}
```
where copyable_move_detector is
```
struct copyable_move_detector {
    constexpr copyable_move_detector()                            = default;
    constexpr copyable_move_detector(const copyable_move_detector&) = default;
    constexpr copyable_move_detector(copyable_move_detector&& rhs) { rhs.been_moved = true; }
    constexpr copyable_move_detector& operator=(const copyable_move_detector&) = default;
    constexpr copyable_move_detector& operator=(copyable_move_detector&& rhs) { rhs.been_moved = true; return *this;}

    bool been_moved = false;
};
```
(not modifying move_detector for fear of affecting something depending on it not being copyable at the moment.)

steve-downey:
Can you merge main back into this branch to resolve conflicts? Otherwise I'll do so from the command line and push.

steve-downey (approved):

jiixyj:
> Can you merge main back into this branch to resolve conflicts? Otherwise I'll do so from the command line and push.

Sure thing, I've just merged the main branch back.

PR #95, opened 2024-12-31
Add missing assignment operator from 'nullopt_t'

jiixyj:
This adds the missing operator and a test.

### DIFF
--- a/src/beman/optional26/tests/optional.t.cpp
+++ b/src/beman/optional26/tests/optional.t.cpp
@@ -868,3 +868,23 @@ TEST(OptionalTest, HashTest) {
         EXPECT_EQ(h1, h2);
     }
 }
+
+TEST(OptionalTest, CanHoldValueOfImmovableType) {
+    struct immovable {
+        explicit immovable()                   = default;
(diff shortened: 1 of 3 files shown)


steve-downey (approved):
:+1:

src/beman/optional26/tests/optional.t.cpp
```diff
@@ -868,3 +868,23 @@ TEST(OptionalTest, HashTest) {
         EXPECT_EQ(h1, h2);
     }
 }
+
+TEST(OptionalTest, CanHoldValueOfImmovableType) {
+    struct immovable {
```
neatudarius, line 873: Please move this test class in https://github.com/bemanproject/optional26/blob/main/src/beman/optional26/tests/test_types.hpp

neatudarius (approved):

src/beman/optional26/tests/optional.t.cpp
jiixyj: Done!

neatudarius (approved):

PR #97, opened 2024-12-31
Set 'engaged' flag on assignment correctly

jiixyj:
Also, refactor some other places to make use of the 'construct' helper that handles the flag.

This adds some tests for the non-trivial assignment operators, for all six cases:

- copy/move from engaged to engaged
- copy/move from engaged to empty
- copy/move from empty to engaged


steve-downey (approved):
:+1:

neatudarius (approved):

PR #99, opened 2024-12-31
Modify constructor constraints to make 'optional<any>' work on libc++

jiixyj:
For `enable_forward_value`, I just reordered the constraints so that the check against `optional` comes first. This fixes a "satisfaction of constraint 'detail::enable_forward_value<T, U>' depends on itself" error on clang.

For `enable_from_other` I introduced a `!std::is_same_v<T, U>` as the first check. This also fixes a "satisfaction of constraint ... depends on itself" error. This check should lead to no change in behavior as in the case where `T` and `U` are the same, `optional<T>`'s copy/move constructors should be used anyway.

### DIFF
--- a/include/beman/optional26/optional.hpp
+++ b/include/beman/optional26/optional.hpp
@@ -202,11 +202,11 @@ concept enable_forward_value = !std::is_same_v<std::decay_t<U>, optional<T>> &&
 
 template <class T, class U, class Other>
 concept enable_from_other =
-    !std::is_same_v<T, U> && std::is_constructible_v<T, Other> && //
-    !std::is_constructible_v<T, optional<U>&> && !std::is_constructible_v<T, optional<U>&&> &&
-    !std::is_constructible_v<T, const optional<U>&> && !std::is_constructible_v<T, const optional<U>&&> &&
-    !std::is_convertible_v<optional<U>&, T> && !std::is_convertible_v<optional<U>&&, T> &&
-    !std::is_convertible_v<const optional<U>&, T> && !std::is_convertible_v<const optional<U>&&, T>;
+    !std::is_same_v<T, U> && std::is_constructible_v<T, Other> && !std::is_constructible_v<T, optional<U>&> &&
+    !std::is_constructible_v<T, optional<U>&&> && !std::is_constructible_v<T, const optional<U>&> &&
+    !std::is_constructible_v<T, const optional<U>&&> && !std::is_convertible_v<optional<U>&, T> &&
+    !std::is_convertible_v<optional<U>&&, T> && !std::is_convertible_v<const optional<U>&, T> &&
+    !std::is_convertible_v<const optional<U>&&, T>;
 
 template <class T, class U>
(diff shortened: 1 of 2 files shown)


steve-downey (commented):
Thanks for these and related PRs! I will take a look at all of them in the next few days.

steve-downey (commented):
I've reproduced this against a few versions of clang/libc++. Older versions fail instead on `!std::is_convertible_v<optional<U>&, T> &&` not being a constant expression, which I'm pretty sure is a bug in libc++. 

Still making sure I understand the changes in enable_from_other. I may have been smarter when I wrote that.

include/beman/optional26/optional.hpp
```diff
@@ -197,16 +197,16 @@ struct hash<optional<T>>;
...
+                               !std::is_same_v<std::decay_t<U>, in_place_t> && std::is_constructible_v<T, U&&>;
 
 template <class T, class U, class Other>
 concept enable_from_other =
-    std::is_constructible_v<T, Other> && !std::is_constructible_v<T, optional<U>&> &&
-    !std::is_constructible_v<T, optional<U>&&> && !std::is_constructible_v<T, const optional<U>&> &&
-    !std::is_constructible_v<T, const optional<U>&&> && !std::is_convertible_v<optional<U>&, T> &&
-    !std::is_convertible_v<optional<U>&&, T> && !std::is_convertible_v<const optional<U>&, T> &&
-    !std::is_convertible_v<const optional<U>&&, T>;
+    !std::is_same_v<T, U> && std::is_constructible_v<T, Other> && //
```
steve-downey, line 205: I don't think the `//` at the end of line needs to be there?

jiixyj: Yes, it doesn't need to be there. Sometimes I do this out of habit to force clang-format to break a line. In this case it makes it so that the next two lines are the `is_constructible_v`s and the two lines after that are the "matching" `is_convertible_v`s.

Should I remove it? The diff would certainly look nicer.

steve-downey: I think removing it will make the diff look a little cleaner. I'm making a note to come back to this to clean up formatting, or refactor into smaller concepts, so it's more obvious what is intended and how they fit together. 

The primary template was really, originally, just to have a base case for the optional<T&> part and I didn't expect to have to revisit it.

jiixyj: OK, I reverted the formatting change and merged main.

steve-downey: Would it be OK to put you in the acknowledgements for the next rev of the paper? It's going out in the next few days and hopefully reviewed and forwarded for C++26 in a few weeks.

jiixyj: > Would it be OK to put you in the acknowledgements for the next rev of the paper? It's going out in the next few days and hopefully reviewed and forwarded for C++26 in a few weeks.

Sure, thank you!

steve-downey (approved):
:+1:

PR #109, opened 2025-02-18
Fixes to optional

steve-downey:
Fixes underlying std::move problem
Changes a static_cast where there is a chance of undefined behavior. 
Paper as presented in Hagenberg.


neatudarius (approved):
LGTM 👍 

Congrats again for optional ref got into C++26!

PR #110, opened 2025-03-02
Fix clang-format issues

ClausKlein:


steve-downey (approved):
LGTM, and I should probably look at getting review dog, or some such set up, as this should have been easier to catch on an earlier PR.

ClausKlein:
> LGTM, and I should probably look at getting review dog, or some such set up, as this should have been easier to catch on an earlier PR.

Something is wrong on CI: this format error was not new: https://github.com/bemanproject/optional/actions/runs/13394192248/job/37408944731

PR #111, opened 2025-03-08
Update CI

steve-downey:
Add  clang 21 to compiler matrix
Update pre-commit from exemplar
Schedule build and test, from exemplar
Update codespell check, was checking papers locally.

### DIFF
--- a/etc/clang-21-toolchain.cmake
+++ b/etc/clang-21-toolchain.cmake
@@ -1,0 +1,11 @@
+# cmake-format: off
+# etc/clang-20-toolchain.cmake -*-cmake-*-
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+# cmake-format: on
+
+include_guard(GLOBAL)
+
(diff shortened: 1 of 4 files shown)


etc/clang-21-toolchain.cmake
```diff
@@ -0,0 +1,11 @@
+# cmake-format: off
+# etc/clang-20-toolchain.cmake -*-cmake-*-
```
neatudarius, line 2:
```suggestion
# etc/clang-21-toolchain.cmake -*-cmake-*-
```

steve-downey: Web sites have trained me not to look at the first few lines of anything, apparently.

neatudarius (approved):
LGTM! Thanks for clang21 suppor!

steve-downey:
> LGTM! Thanks for clang21 suppor!

Sadly, GCC15 is only available easily on the beta Ubuntu, Plucky. I'll squash out and push the attempts to get it working.

PR #113, opened 2025-03-13
Find package

steve-downey:
Use the new find-package support from the exemplar project, add documentation, and make the failure to find the GTest package in the common case a little friendlier.

### DIFF
--- a/.markdownlint.yaml
+++ b/.markdownlint.yaml
@@ -2,9 +2,6 @@
 # Disable inline html linter is needed for <details> <summary>
 MD033: false
 
-# MD013/line-length : Line length : https://github.com/DavidAnson/markdownlint/blob/v0.35.0/doc/md013.md
-MD013: false
-
 # MD024/no-duplicate-heading : https://github.com/DavidAnson/markdownlint/blob/main/doc/md024.md
 # Supress warning about the same heading twice unless they are in the same block
 MD024:
--- a/.markdownlint.yaml
+++ b/.markdownlint.yaml
@@ -3,10 +3,7 @@
 MD033: false
 
 # MD013/line-length : Line length : https://github.com/DavidAnson/markdownlint/blob/v0.35.0/doc/md013.md
-# Conforms to .clang-format ColumnLimit
-# Update the comment in .clang-format if we no-longer tie these two column limits.
-MD013:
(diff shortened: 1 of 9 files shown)


.markdownlint.yaml
```diff
@@ -2,12 +2,6 @@
 # Disable inline html linter is needed for <details> <summary>
 MD033: false
 
-# MD013/line-length : Line length : https://github.com/DavidAnson/markdownlint/blob/v0.35.0/doc/md013.md
-# Conforms to .clang-format ColumnLimit
-# Update the comment in .clang-format if we no-longer tie these two column limits.
-MD013:
```
neatudarius, line 8: Should we also set `MD013: false` here? I noticed you set it on `.pre-commit-config.yaml`.

tests/beman/optional/CMakeLists.txt
```diff
@@ -31,12 +26,12 @@ target_link_libraries(
     beman_optional_test
     PRIVATE beman_optional GTest::gtest GTest::gtest_main
 )
-
 # Issue #32: Re-enable ASAN run CI/clang-19.
 #
 # Note: clang-19 + gtest_discover_tests + Asan setup causes errors on some
 # platforms. Temporary switch to gtest_add_tests and skip some Asan checks.
 # Change also applied for CI flows.
+include(GoogleTest)
```
neatudarius, line 34: I think it was here, and during some review I got feedback from multiple people to put all includes at the top of the file.

neatudarius (approved):

.markdownlint.yaml
steve-downey: As near as I've been able to figure out `MD013: false` doesn't actually work, it still ends up with the default line length of 80 being checked. The property isn't boolean, like the ones we have set to `false`.

tests/beman/optional/CMakeLists.txt
steve-downey: I've seen it both ways. You can't factor out CML.txt into distinct components, though, so spreading out includes across the file tends to mean you accumulate includes with no idea which uses what. You get one CML.txt per directory so the more effective maintainable style on projects I've been on have been to keep related things together. 
CMake is not C++.

PR #117, opened 2025-04-04
Lwg 20250314

steve-downey:


examples/base_derived_cast.cpp
```diff
@@ -0,0 +1,25 @@
+// examples/optional_ref.cpp -*-C++-*-
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
```
neatudarius, line 2: Thanks for splitting these examples!

lwg-20250314.org
```diff
@@ -0,0 +1,102 @@
...
+#+subtitle:
+#+infojs_opt:
+#+creator: <a href="https://www.gnu.org/software/emacs/">Emacs</a> 30.1.50 (<a href="https://orgmode.org">Org</a> mode 9.7.11)
+#+latex_header:
+
+* Notes
+https://wg21.link/P2988r9 std::optional<T&>
+SD presenting
+[optionalref.iterators]
+TS: p1: “These types” -> “This type”
```
neatudarius, line 37: Can you please format them?

image


L.E. Why do we have duplicated notes?

PR #118, opened 2025-04-04
Merge results of LWG review

steve-downey:
The optional header is in sync with the specification of std::optional. Certain test cases have been added for detecting ambiguous conversions that optional should resolve. These ambiguities were introduced when converting from `if else` constructs to the ternary `? :` operator, as that has rules for calculating a common type, where if / else does not.

- initialization of value_ is via utility function
- value_or supports braced initialization
- value_or decays the copy it makes, removing cv qualifiers
- borrowed range flag is in a more direct style
- value_ is initialized to nullptr, the nullopt constructor delegates, and other constructors no longer branch

Paper is post LWG review and is immutable at P2988R12.


examples/optional_ref.cpp
```diff
@@ -44,36 +44,15 @@ beman::optional::optional<Cat&> api() {
...
-int example1() {
+int example() {
     // Example from P2988R5: optional reference.
     [[maybe_unused]] Cat*                            old_cat = std17::api();
     [[maybe_unused]] beman::optional::optional<Cat&> new_cat = std26::api();
 
     return 0;
 }
 
-struct derived;
```
neatudarius, line 55: Thanks for splitting this!

neatudarius (approved):
LGTM and good job!

PS Didn't review content under `papers/`.

neatudarius:
Should we merge this?

steve-downey:
> Should we merge this?

Yes. I was in Scotland last week, and had limited internet.

PR #124, opened 2025-05-26
Format kind

steve-downey:
GCC 15 release branch and GCC 16 have support for format_kind, revealing that the support in beman.optional has always been broken.


neatudarius (approved):

steve-downey:
Accidentally deleted the branch in my fork. 
GCC 14 failure is fixed in separate commit which I'll merge first and then this,

PR #125, opened 2025-05-26
More coverage

steve-downey:
Add more coverage. 
May make coverage percent go down as more functions are instantiated.


neatudarius (approved):

steve-downey:
Accidentally deleted the branch in my fork. 
GCC 14 failure is fixed in separate commit which I'll merge first and then.

PR #193, opened 2026-03-17
beman-submodule update

ClausKlein:
This update the beman-submodule `infra`

And do the needed changes on build scripts.


steve-downey:
Closing after incorporating this into #194

